PR #17, opened 2025-06-27
Add an accessor to get the selection

AdamFelt:


include/hvt/engine/viewportEngine.h
```diff
@@ -103,8 +103,8 @@ struct HVT_API FramePassDescriptor
     /// Default unique identifier (which can be customized when the render index is shared).
     PXR_NS::SdfPath uid;
 
-    /// Paths to exclude by render tasks.
-    PXR_NS::SdfPathVector excludedPaths; // None by default.
+    /// Light paths to exclude by render tasks.
+    PXR_NS::SdfPathVector excludedLightPaths; // None by default.
```
AdamFelt, line 107: Renamed this because it was confusing.  It is not the exclude list for the objects rendered in the frame pass, it is the light exclude list (aka light linking)

source/engine/selectionHelper.cpp
```diff
@@ -48,6 +48,14 @@ void SelectionHelper::SetSelection(HdSelectionSharedPtr selection)
...
 }
 
+SdfPathVector SelectionHelper::GetSelection(PXR_NS::HdSelection::HighlightMode highlightMode) const
+{
+    HdSelectionSharedPtr selection = _selectionTracker->GetSelectionMap();
+    if (selection)
+        return selection->GetSelectedPrimPaths(highlightMode);
+    return { };
+}
+
```
AdamFelt, line 58: New function to get the selection sets for the render pass.  Used to logically pass to other HdTasks who may want to treat a set of objects differently, but may not explicitly want to assume the set is the selection (from the render context).

sebastienberube-adsk (approved):
I am approving, BUT make sure to update any affected file in Components or HVT that might refer to the renamed `FramePassDescriptor::excludedLightPaths` variable.

source/engine/selectionHelper.cpp
pixnblox: Could this comment (or similar) be added as a `\note` to the header file? (to share how this could be used)

PR #22, opened 2025-07-04
OGSMOD-7567 - Prep. work for the HVT CI/CD

hodoulp:
The pull request disables one unit test when compiling with OpenUSD `origin/dev` because it needs a `adsk/dev` feature to succeed.


erikaharrison-adsk (approved):
I think we should change to using PXR versioning and not the ADSK_PENDING_ as I don't think that's accurate. Please adjust. Otherwise minor comment.

Have confirmed via internal CI that these pass - thank you!

erikaharrison-adsk (approved):
Looks good, thanks for tackling this!

PR #23, opened 2025-07-04
OGSMOD-7666 - Try to reproduce the issue

hodoulp:
The pull request tries to reproduce the problem reported by MayaHydra using a unit tests. However, it works fine in that context.

### DIFF
--- a/test/tests/testMultiSampling.cpp
+++ b/test/tests/testMultiSampling.cpp
@@ -1,0 +1,298 @@
+// Copyright 2025 Autodesk, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


test/tests/testMultiSampling.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+            params.colorspace      = pxr::HdxColorCorrectionTokens->disabled;
+            params.backgroundColor = TestHelpers::ColorDarkGrey;
+            params.selectionColor  = TestHelpers::ColorYellow;
+
+            // Do not display right now, wait for the second frame pass.
+            params.enablePresentation = false;
+
+            // That's the default value, but enforce it to true in case...
+            params.enableMultisampling = true;
```
hodoulp, line 145: This is the default. With or without this line, the result seems right.

PR #29, opened 2025-07-15
Disable the selection provider test on android for now to avoid a crash.

AdamFelt:


erikaharrison-adsk (approved):
Currently it seems to be causing crashes in Android tests. Requires more investigation.

PR #30, opened 2025-07-16
OGSMOD-7577: Add support for Neye

crescentyue74:
This makes the "Neye" (view-space normal) AOV available to users of HVT.

### DIFF
--- a/source/engine/framePass.cpp
+++ b/source/engine/framePass.cpp
@@ -292,9 +292,9 @@ HdTaskSharedPtrVector FramePass::GetRenderTasks(RenderBufferBindings const& inpu
     {
         if (!IsStormRenderDelegate(GetRenderIndex()) || params().enableOutline)
             renderOutputs = { HdAovTokens->color, HdAovTokens->depth, HdAovTokens->primId,
-                HdAovTokens->elementId, HdAovTokens->instanceId };
+                HdAovTokens->elementId, HdAovTokens->instanceId, HdAovTokens->Neye };
         else
-            renderOutputs = { HdAovTokens->color, HdAovTokens->depth };
+            renderOutputs = { HdAovTokens->color, HdAovTokens->depth, HdAovTokens->Neye };
     }
 
     _bufferManager->SetRenderOutputs(renderOutputs, inputAOVs, {});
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -184,7 +184,7 @@ endif()
 # This section only adds settings applicable to all the projects.
 
 # Set the HVT include (public API) and source (private API) directories, for use by sub-projects.
-set(_HVT_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include")
+set(_HVT_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/include" CACHE PATH "HVT include directory")
```
hodoulp, line 187: In theory this should never be changed, and the under score means that the variable is private. Why do you need this change?

source/engine/framePass.cpp
```diff
@@ -294,7 +294,7 @@ HdTaskSharedPtrVector FramePass::GetRenderTasks(RenderBufferBindings const& inpu
             renderOutputs = { HdAovTokens->color, HdAovTokens->depth, HdAovTokens->primId,
                 HdAovTokens->elementId, HdAovTokens->instanceId };
         else
-            renderOutputs = { HdAovTokens->color, HdAovTokens->depth };
+            renderOutputs = { HdAovTokens->color, HdAovTokens->depth, HdAovTokens->Neye };
```
hodoulp, line 297: Why not changing the line 294?

hodoulp (commented):
Everything seems right. Only have one question.

source/engine/framePass.cpp
crescentyue74: According to https://forum.aousd.org/t/pickfromrenderbuffer-missing-normal-data/1354 and current USD code, I think adding `HdAovTokens->Neye` won't work for case `!IsStormRenderDelegate(GetRenderIndex())`.

CMakeLists.txt
crescentyue74: I need to use one of the header in hvt.
Was having trouble to include that file but now I found the proper path to use. Removed this change.

hodoulp (approved):

source/engine/framePass.cpp
sebastienberube-adsk: @crescentyue74 Is there a performance cost associated with the addition of this new renderOutput, and if there could be one, what is this cost? I believe this will add an additional texture attachment to all the drawing passes.

There are a lot of AOV buffers we can request here, such as primId with Storm, but it comes with a cost.

I am wondering if we could make this Neye render ouput optional, in the [FramePassParams](https://github.com/Autodesk/hydra-viewport-toolbox/blob/f23892648b673fbfde7f1cb75fcc6a26ed89e323/include/hvt/engine/framePass.h#L117-L140).

PR #31, opened 2025-07-17
As we move DashDotLines to a separate rprim, we will not build the BasisCurves with style.

PierreWang:
As we move DashDotLines to a separate rprim, we will not build the BasisCurves with style.


hodoulp (approved):
LGMT

PR #32, opened 2025-07-18
OGSMOD-7744 - Fix utest wrong mgt of shared buffers

hodoulp:
The utest is only displaying the `depth` so there is no need to request for the color buffer (which doesn't exist) and also, no need to use `hvt::CreateRenderBufferProxy()` (dedicated to share AOV buffers between arbitrary render delegate and Storm).


erikaharrison-adsk (approved):
Seems fine, but I'm not sure where to confirm it works for all platforms.

PR #37, opened 2025-07-25
AGPHVT-40: WIP.

sm-adsk:


include/hvt/engine/framePass.h
```diff
@@ -127,7 +127,7 @@ struct HVT_API FramePassParams : public BasicLayerParams
     bool enableColorCorrection { true };
     PXR_NS::GfVec4f backgroundColor { 0.025f, 0.025f, 0.025f, 1.0f };
     float backgroundDepth {1.0f};
-    bool clearBackground{ true };
+    bool clearBackgroundColor{ true };
```
hodoulp, line 130: This name is right but ambiguous, add comment to explain the tow variables  `clearBackgroundColor` & `clearBackgroundDepth`. It could be:
`/// Clear the background of the color buffer.`
`/// Clear the background of the depth buffer.`

source/engine/framePass.cpp
```diff
@@ -317,16 +317,30 @@ HdTaskSharedPtrVector FramePass::GetRenderTasks(RenderBufferBindings const& inpu
...
+                HdAovTokens->color, VtValue(_passParams.backgroundColor));
+        }
+        else
+        {
+            _bufferManager->SetRenderOutputClearColor(HdAovTokens->color, VtValue());
+        }
     }
 
-    if (_passParams.clearBackgroundDepth)
+    if (_passParams.visualizeAOV == HdAovTokens->depth)
```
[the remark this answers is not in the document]
hodoulp: We now understand the fix and it's consistent between the `color` and `depth` buffers.

source/engine/renderBufferManager.cpp
```diff
@@ -485,8 +485,16 @@ void RenderBufferManager::Impl::SetRenderOutputClearColor(
     }
 
     // Check if we're setting a value for a nonexistent AOV.
-    SdfPath renderBufferId                          = GetAovPath(controllerId, name);
-    _aovTaskCache.outputClearValues[renderBufferId] = clearValue;
+    SdfPath renderBufferId = GetAovPath(controllerId, name);
```
hodoulp, line 488: it must be a const temporary variable.

test/tests/testFramePasses.cpp
```diff
@@ -756,9 +756,140 @@ TEST(TestViewportToolbox, TestFramePasses_TestDynamicAovInputs)
...
+    ASSERT_TRUE(context->_backend->saveImage(imageFile));
+
+    ASSERT_TRUE(context->_backend->compareImages(imageFile));
+}
+// Note: The second frame pass is not displayed on Android. Refer to OGSMOD-7277.
+// Note: The two frame passes are displayed in the left part on iOS. Refer to OGSMOD-7278.
+#if defined(__ANDROID__) || TARGET_OS_IPHONE == 1
+TEST(TestViewportToolbox, DISABLED_TestFramePasses_MultiViewportsFramePassClearBufferImprovements)
+#else
+TEST(TestViewportToolbox, TestFramePasses_MultiViewportsFramePassClearBufferImprovements)
```
hodoulp, line 790: Following the previous changes, there should be a unit test to validate that the clear background also works for the color buffer.

test/tests/testFramePasses.cpp
```diff
@@ -756,9 +756,140 @@ TEST(TestViewportToolbox, TestFramePasses_TestDynamicAovInputs)
...
+            params.viewInfo.lights           = stage2.defaultLights();
+            params.viewInfo.material         = stage2.defaultMaterial();
+            params.viewInfo.ambient          = stage2.defaultAmbient();
+
+            params.colorspace = HdxColorCorrectionTokens->disabled;
+
+            // Clear color and depth for the first 5 frames, then stop
+            // clearing for the final render (after frame 5). This will validate
+            // the clear does not "stick" once it is enabled.
+            params.clearBackgroundColor      = (frameCount > 5);
```
hodoulp, line 885: As it only displays the depth buffer, is the clear of the color buffer needed?

test/tests/testFramePasses.cpp
```diff
@@ -756,9 +756,140 @@ TEST(TestViewportToolbox, TestFramePasses_TestDynamicAovInputs)
...
+            params.clearBackgroundDepth = true;
+            params.backgroundColor      = TestHelpers::ColorDarkGrey;
+            params.backgroundDepth      = 1.0f;
+            params.selectionColor       = TestHelpers::ColorYellow;
+
+            // Only visualizes the depth.
+            params.visualizeAOV = HdAovTokens->depth;
+
+            // Displays the depth aov.
+            params.enablePresentation = true;
```
hodoulp, line 852: Oops it must be false.

test/tests/testFramePasses.cpp
```diff
@@ -756,15 +756,283 @@ TEST(TestViewportToolbox, TestFramePasses_TestDynamicAovInputs)
...
+    ASSERT_TRUE(context->_backend->compareImages(imageFile));
+}
+
+// Note: The second frame pass is not displayed on Android. Refer to OGSMOD-7277.
+// Note: The two frame passes are displayed in the left part on iOS. Refer to OGSMOD-7278.
+#if defined(__ANDROID__) || TARGET_OS_IPHONE == 1
+TEST(TestViewportToolbox,
+    DISABLED_TestFramePasses_MultiViewportsFramePassClearColorBufferImprovements)
+#else
+TEST(TestViewportToolbox, TestFramePasses_MultiViewportsFramePassClearColorBufferImprovements)
```
hodoulp, line 925: The test is only about clearing the color buffer, you can shorten the test name (and as dise-effect the baseline image name) e.g., `TestFramePasses_ClearColorBuffer`

test/tests/testFramePasses.cpp
```diff
@@ -756,15 +756,283 @@ TEST(TestViewportToolbox, TestFramePasses_TestDynamicAovInputs)
...
+            // Only visualizes the color.
+            params.visualizeAOV = HdAovTokens->color;
+
             // Gets the list of tasks to render but use the render buffers from the first frame
             // pass.
             const HdTaskSharedPtrVector renderTasks =
                 framePass2.sceneFramePass->GetRenderTasks(inputAOVs);
 
+            // Displays the color aov.
+            params.enablePresentation = true;
```
hodoulp, line 1034: ⚠️ All the parameter updates must be done before requesting the list of tasks. This call (i.e., `GetRenderTasks()`) updates all the task with the parameter values.

test/tests/testFramePasses.cpp
```diff
@@ -756,15 +756,283 @@ TEST(TestViewportToolbox, TestFramePasses_TestDynamicAovInputs)
...
+            // Only visualizes the depth.
+            params.visualizeAOV = HdAovTokens->depth;
+
+            // Gets the list of tasks to render but use the render buffers from the first frame
+            // pass.
+            const HdTaskSharedPtrVector renderTasks =
+                framePass2.sceneFramePass->GetRenderTasks(inputAOVs);
+
+            // Displays the depth aov.
+            params.enablePresentation = true;
```
hodoulp, line 899: ⚠️ All the parameter updates must be done before requesting the list of tasks. This call (i.e., `GetRenderTasks()`) updates all the task with the parameter values.

hodoulp (approved):

source/engine/framePass.cpp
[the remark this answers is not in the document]
sebastienberube-adsk: This looks incorrect, I agree with Adam. I just catched this while testing rendering with multiple frame passes and realized the depth clear flag was ignored. We should be able to control if we want to clear the depth or not, regardless if the depth is currently being visualized or not.

> [hodoulp](https://github.com/hodoulp) [last month](https://github.com/Autodesk/hydra-viewport-toolbox/pull/37/files#r2240465760)
> We now understand the fix and it's consistent between the color and depth buffers.

Regarding the consistency with the color, maybe the reason why the clear color was not set if the currently visualized AOV is not color might be because we won't see any difference in this case. But this is different with the depth, as it will impact what we see even if we don't visualize depth. Not sure, just trying to find some logic in this old code.

PR #38, opened 2025-07-25
Fix a former merge conflict

hodoulp:
The parameter was duplicated adding confusion in the API.


sm-adsk (approved):

sebastienberube-adsk (approved):
As the parameter was duplicated, I approve its removal. (it also existing in [ModelParams](https://github.com/Autodesk/hydra-viewport-toolbox/blob/673f052faa708741f8e0f4fa2004d5362d286296/include/hvt/engine/framePass.h#L71) `FramePassParams::modelInfo`)

PR #41, opened 2025-07-25
OGSMOD-7803 - MayaHydra - Add framing to the public API

hodoulp:
The pull request adds the framing in the public API and unit tests to demonstrate that it works.

In fact, it replaces the former `ViewportportRect` by the generic `pxr::CameraUtilFraming` introducing lot of changes. I still picked that solution to have the most generic method directly in the public API. 

Note: There are some changes in all the unit tests (explained the number of changed files in this pull request). But these changes are quite simple and there are two helpers to make that transition even easier.

### DIFF
--- a/test/tests/testFramePasses.cpp
+++ b/test/tests/testFramePasses.cpp
@@ -1101,9 +1101,13 @@ TEST(TestViewportToolbox, TestFramePasses_DisplayClipping1)
 
         params.renderBufferSize = pxr::GfVec2i(width, height);
         // Takes all the rendered image but only displays the left part.
-        params.viewInfo.framing = { { { 0, 0 },
-                                  { static_cast<float>(width), static_cast<float>(height) } },
-                { { 0, 0 }, { width / 2, height } }, 1.0f };
+        params.viewInfo.framing = {
+            // Data window: full render buffer.
+            { { 0, 0 }, { static_cast<float>(width), static_cast<float>(height) } },
+            // Display window: left part only.
+            { { 0, 0 }, { width / 2, height } }, 
+            1.0f
+        };
 
         params.viewInfo.viewMatrix       = stage.viewMatrix();
         params.viewInfo.projectionMatrix = stage.projectionMatrix();


hodoulp:
> I tested it and it does work for our purposes. However, I am wondering if there is a benefit to having the framing param be part of the ViewParams/viewInfo, when it is simply used to set renderParams.framing, which is already accessible?

My point is that it make the API more understandable / readable i.e., set exactly what you need.

hodoulp:
Before moving forward, can @AdamFelt and/or @pixnblox have a look?

hodoulp:
cc'ing @pixnblox, @erikaharrison-adsk and/or @sebastienberube-adsk can you have a look?

test/tests/testFramePasses.cpp
```diff
@@ -1050,3 +1074,130 @@ TEST(TestViewportToolbox, TestFramePasses_ClearColorBuffer)
 
     ASSERT_TRUE(context->_backend->compareImages(imageFile));
 }
+
+TEST(TestViewportToolbox, TestFramePasses_DisplayClipping1)
```
kohakukun, line 1078: If I understand correctly, this is the use case which was not possible with only the viewport?

test/tests/composeTaskHelpers.cpp
```diff
@@ -72,8 +72,10 @@ void RenderFirstFramePass(TestHelpers::FramePassInstance& framePass1, int width,
 {
     hvt::FramePassParams& params = framePass1.sceneFramePass->params();
 
-    params.renderBufferSize          = GfVec2i(width, height);
-    params.viewInfo.viewport         = { { 0, 0 }, { width, height } };
+    params.renderBufferSize = GfVec2i(width, height);
```
kohakukun, line 75: Nit: I would use the style that was used before and is being used in the other lines. With the spaces to generate alignment. This also will reduce the list of changes

include/hvt/engine/framePass.h
```diff
@@ -304,10 +319,10 @@ class HVT_API FramePass
     /// \return A collection of parameters that can be set for this frame pass.
     inline const FramePassParams& params() const { return _passParams; }
 
-    /// Gets the viewport dimensions.
-    virtual const PXR_NS::GfVec4i GetViewport() const
+    /// Gets the viewport positions & dimensions.
+    virtual const PXR_NS::GfRange2f GetViewport() const
```
kohakukun, line 323: IMO since we are moving away from the viewport keyword, we might change the getter as well and have maybe `GetDisplayWindow` and `GetDataWindow`

kohakukun (commented):
LGTM in general but, as this is open source, we might want to keep a changelog since this is a breaking change. This change would increase the major version

test/tests/testFramePasses.cpp
```diff
@@ -1050,3 +1074,130 @@ TEST(TestViewportToolbox, TestFramePasses_ClearColorBuffer)
...
+    auto context = TestHelpers::CreateTestContext();
+
+    TestHelpers::TestStage stage(context->_backend);
+    ASSERT_TRUE(stage.open(context->_sceneFilepath));
+
+    TestHelpers::FramePassInstance framePass =
+        TestHelpers::FramePassInstance::CreateInstance(stage.stage(), context->_backend);
+
+    // Render 10 times (i.e., arbitrary number to guaranty best result).
+    int frameCount = 10;
```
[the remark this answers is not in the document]
hodoulp: Sometime you need several passes to have a 'complete' render when using `Storm`.

include/hvt/engine/framePass.h
hodoulp: Lot of impacts on existing code but it's a good idea.

kohakukun (approved):

test/tests/testFramePasses.cpp
```diff
@@ -1052,3 +1076,130 @@ TEST(TestViewportToolbox, TestFramePasses_ClearColorBuffer)
...
+        hvt::FramePassParams& params = framePass.sceneFramePass->params();
+
+        const auto width = context->width();
+        const auto height = context->height();
+
+        params.renderBufferSize = pxr::GfVec2i(width, height);
+        // Takes all the rendered image but only displays the left part.
+        params.viewInfo.framing = { { { 0, 0 },
+                                  { static_cast<float>(width), static_cast<float>(height) } },
+                { { 0, 0 }, { width / 2, height } }, 1.0f };
```
sebastienberube-adsk, line 1106: [Minor]

I prefer the comment style that was used for TestFramePasses_DisplayClipping2:
```
// Data window: full render buffer.
{ { 0, 0 }, { static_cast<float>(width), static_cast<float>(height) } },
// Display window: left part only
{ { 0, 0 }, { width / 2, height } }, 1.0f };
```

include/hvt/engine/framePass.h
```diff
@@ -88,9 +87,24 @@ struct HVT_API ViewParams
...
+    }
+
+    /// Helper to get a default framing.
+    static PXR_NS::CameraUtilFraming GetDefaultFraming(int posX, int posY, int width, int height)
+    {
+        /// \note This is to display all the render buffer content into the screen potentially
+        /// moving its origin and resizing it.
+        return PXR_NS::CameraUtilFraming(
+            PXR_NS::GfRect2i(PXR_NS::GfVec2i(posX, posY), width, height));
+    }
```
sebastienberube-adsk, line 107: [Minor]

That is subjective, but I would personally move these 2 static helpers out of the ViewParams struct:
- GetDefaultFraming(int width, int height)
- GetDefaultFraming(int posX, int posY, int width, int height)

sebastienberube-adsk (approved):

PR #42, opened 2025-07-28
Enable virutal GPU for tests

laforel:
Implements comprehensive virtual GPU support for running GPU-dependent tests in headless GitHub Actions Linux runners using Xvfb and Mesa software rendering.

- Added Xvfb (X Virtual Framebuffer) setup for headless OpenGL rendering
- Configured Mesa LLVMpipe software renderer with optimized settings
- Set up proper environment variables (DISPLAY, LIBGL_ALWAYS_SOFTWARE, GALLIUM_DRIVER)

- Fixed Linux RPATH configuration (`$ORIGIN/` vs macOS `@loader_path/`)
- Added `--disable-new-dtags` linker flags to force RPATH over RUNPATH
- Ensured automatic OpenUSD library discovery without manual LD_LIBRARY_PATH

- Created Linux-specific baseline images for all rendering tests
- Consolidated platform-specific baseline strategy using `_linux` suffix
- Added support for new tests: `TestFramePasses_ClearDepthBuffer` and `TestFramePasses_ClearColorBuffer`

- Updated `TestSearchEdges` and `TestSearchPoints` with Linux-specific expected results
- Fixed algorithmic differences between platforms in search functionality

- Limited test execution to Linux runners only (unlimited resources)
- Cleaned up redundant environment variable declarations
- Maintained platform-agnostic CMake presets


hodoulp:
cc'ing @erikaharrison-adsk In order to use the same baseline images for Linux (software vs. hardware) & thinking that we only want to trap major errors can we investigate the idea of having a different threshold for swf Linux?
So we can use the existing Linux baseline images while waiting for the right long-term solution.

PR #46, opened 2025-08-01
[Draft] OGSMOD-7842: Remove nEye from AovInputTask and from RenderBufferManager

sebastienberube-adsk:
nEye buffer can be accessed from the RenderIndex instead of the TaskContext.


hodoulp:
👍 Waiting for the final pull request but approved what I see.

PR #48, opened 2025-08-06
NOM-3927 - Fix FXAATask so it works with non-square viewports

fanciful-marmot:
Prior to this commit, it looks like a single float was used to scale UVs to try to find neighbouring pixels based on this section

```glsl
    // Sample current and neighboring pixels luminance in axis aligned space.
    vec4 extent;
    extent.xy = uvOut + vec2(-0.5, -0.5) * uResolution;
    extent.zw = uvOut + vec2( 0.5,  0.5) * uResolution;
```

Presumably for FXAA to work you want to find the direct neighbours of the pixel you're working on to decide how to blur. Depending on the actual resolution, `uResolution` would have to changed if that's the case. For example, at a resolution of 100x100, you'd want the delta UV to be +/- (0.01, 0.01) to find the direct neighbours so uResolution would have to be 0.02.
However, if the resolution is 1000x1000, you'd want the delta UV to be +/- (0.001, 0.001) to find the direct neighbours so uResolution would have to be 0.002. If you'd used the same uResolution you had used for the 100x100 example, you'd be sampling 10 pixels away which doesn't seem right.

So `uResolution` should be adjusted based on the actual resolution. This leads to the next problem of it being a single float. If you have a non-square aspect ratio you'd get the direct neighbours for one dimension but be off for the other.

This PR changes `uResolution` to `vec2` instead of `float` and updates the examples to demonstrate how this value should be set.

### NOTE
I changed the example and the test to not overdo the effect. We can make it overdo it by setting using a number > 1 in `1.0f / framing.dataWindow.GetWidth()` if we want.

Also the tests that do run will fail until I update the baseline images. I'll need to let it fail once to get it for all platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,9 @@ if (MSVC)
     add_compile_options(/wd4251)  # DLL interface warnings.
     add_compile_options(/wd4275)  # DLL interface warnings.
 
+    # Enable C++ exception handling (required for older MSVC versions < 19.44, harmless for newer versions).
+    add_compile_options(/EHsc)
+
     if (ENABLE_WARNINGS_AS_ERRORS)
         add_compile_options(/WX)
     endif()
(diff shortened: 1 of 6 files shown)


source/tasks/fxaaTask.cpp
```diff
@@ -81,11 +81,11 @@ void FXAATask::_Sync(HdSceneDelegate* delegate, HdTaskContext* /* ctx */, HdDirt
         HgiShaderFunctionAddStageInput(&shaderDesc, "uvOut", "vec2");
         HgiShaderFunctionAddTexture(&shaderDesc, "colorIn", 0);
         HgiShaderFunctionAddStageOutput(&shaderDesc, "hd_FragColor", "vec4", "color");
-        HgiShaderFunctionAddConstantParam(&shaderDesc, "uResolution", "float");
+        HgiShaderFunctionAddConstantParam(&shaderDesc, "uResolution", "vec2");
```
erikaharrison-adsk, line 84: Minor: Is this now uvResolution?

test/tests/testFramePasses.cpp
```diff
@@ -257,9 +254,14 @@ TEST(TestViewportToolbox, TestFramePasses_MainWithFxaa)
...
             {
+                auto framing = _sceneFramePass->params().renderParams.framing;
+
                 const VtValue value        = fnGetValue(HdTokens->params);
                 hvt::FXAATaskParams params = value.Get<hvt::FXAATaskParams>();
-                params.resolution          = fxaaResolution;
+                params.pixelToUV           = GfVec2f(
+                    1.0f / framing.dataWindow.GetWidth(),
+                    1.0f / framing.dataWindow.GetHeight()
+                );;
```
erikaharrison-adsk, line 264: Super minor: two semi colons.

erikaharrison-adsk (approved):
Thanks for this fix! Looks good to me, using width/height instead of a shared fixed value. May want to ping Jochen directly as he worked more closely on this.

source/tasks/fxaaTask.cpp
fanciful-marmot: I took the `u` part to mean `uniform`, not `u` as in `uv` 😅

hodoulp:
Are the values used to test `FXAA` produce a baseline image different enough to compare with a utest without FXAA?

PR #52, opened 2025-08-16
Configure include directory propagation in submodule scenarios

laforel:
This fix was found after Jason Bellenger with Nomad client reported HVT’s include path missing while using HVTEx as submodule in his project.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -91,6 +91,13 @@ target_link_libraries(${_TARGET}
     PRIVATE ${_PRIVATE_LIBRARIES}
 )
 
+# Set the required include directories for consumers of hvt
+target_include_directories(${_TARGET}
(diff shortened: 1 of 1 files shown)


source/geometry/geometry.cpp
```diff
@@ -46,7 +46,7 @@
 #include <pxr/imaging/hdSt/material.h>
 #include <pxr/imaging/hdSt/materialNetwork.h>
 #include <pxr/imaging/hio/glslfx.h>
-#include <pxr/imaging/pxOsd/tokens.h>
+//#include <pxr/imaging/pxOsd/tokens.h>
```
hodoulp, line 49: Remove the line

hodoulp (changes requested):

hodoulp (approved):

PR #55, opened 2025-08-22
OGSMOD-7777: allow clipping on main frame pass

muenstc:
- Adds the capability to pass a list of clip planes to the the frame pass via the view parameters.
This allows to use the hardware clipping feature of hdstorm.
- add rotation parameter to dataSource interface
- UpdatePrim now handles rotation for xformOp:orient and allows visibilityOverrides

### DIFF
--- a/include/hvt/resources/gizmos/moveManipulator.usda
+++ b/include/hvt/resources/gizmos/moveManipulator.usda
@@ -242,6 +242,7 @@ def Xform "moveManipulator"
     }
     def Xform "sectionPlane"
     {
+        token visibility = "invisible"
         def "plane" (prepend specializes = </__class/basePlane>; prepend apiSchemas = ["MaterialBindingAPI"])
         {
             token axis = "Y"


source/engine/viewportEngine.cpp
```diff
@@ -812,28 +817,61 @@ void UpdatePrim(UsdStageRefPtr& stage, const SdfPath& path, const GfVec3d& posit
...
+            for (const auto& childPrim : descendants)
+            {
+                UsdGeomImageable(childPrim).MakeVisible();
+            }
+            
+            // Then apply specific visibility overrides from the map
+            for (const auto& [primName, childVisible] : childVisibility)
+            {
+                // Construct child path using string concatenation to avoid TfToken creation
+                SdfPath childPath(path.GetString() + "/" + primName);
```
sebastienberube-adsk, line 866: SdfPaths are efficient for quick comparison once they are created, but they are expensive to generate.

I suggest changing the type of `childVisibility` from:
`const std::map<std::string, bool>&`
 to
 `const std::map<SdfPath, bool>&`
 
This will make it possible for the caller to create the SdfPath once and avoid sting manipulations in this UpdatePrim function.

I know it won't make an actual performance difference for the manipulators; the main idea is to set a good SdfPath usage example, as the SdfPath class is known to have caused performance issues when used in loops without caching (they are really meant to be created once and reused).

include/hvt/engine/framePass.h
```diff
@@ -110,6 +111,14 @@ struct HVT_API ViewParams
     PXR_NS::GfVec4f ambient { PXR_NS::GfVec4f(0.0f, 0.0f, 0.0f, 0.0f) };
 
     /// @}
+
+    /// \name Setup around section planes for clipping.
+    /// @{
+
+    /// Section planes for clipping using GfPlane objects
+    std::vector<PXR_NS::GfPlane> sectionPlanes;
```
hodoulp, line 119: ⚠️ Why this attribute must be in such a generic frame pass?
Most of the frame passes will never 'use' the section plane i.e., all the secondary frame passes for example!

PR #57, opened 2025-08-29
OGSMOD-7977 - Cleanup & fix cmake files 

hodoulp:
The pull request cleans up and fixes some cmake files.
The main improvement is that it now generates only one binary i.e., the `hvt` library (static or dynamic).


hodoulp:
The `Components` and `HVTEx` pull requests are now green.

PR #60, opened 2025-09-08
OGSMOD-7969: Fix Frame Pass Depth Clear Flag

sebastienberube-adsk:
OGSMOD-7969 Fix FramePass Depth Clear Flag implementation

The render buffer clear flag was ignored, unless the render buffer was visualized.
This logic was wrong, as it is totally acceptable to control whether or not the depth or color buffer are cleared without having to visualize them. This is especially important for the depth buffer, which is typically NOT visualized.

The clear logic was also simplified by removing the call to ResetRenderOutputClear() from RenderBufferManager::Impl::SetRenderOutputs(). This was the old way to clear the render buffers and this function is no more necessary, since RenderBufferManager::SetRenderOutputClearColor() can now be called with an empty VtValue instead, which will yield the same result and reset the clear value for a given AOV name. Furthermore, it is preferable to set the clear value outside of RenderBufferManager::Impl::SetRenderOutputs(), which already contains too much complexity, and which would also early out in some cases, also failing to clear the render outputs. For these reasons, ResetRenderOutputClear() was removed.

The unit test weren't updated since there are already 2 good test cases for clearing the depth:
- TestViewportToolbox.TestFramePasses_ClearDepthBuffer
- TestViewportToolbox.TestFramePasses_ClearColorBuffer

The only reason why the depth clear bug wasn't catched by these tests is because they are configured to visualize the same AOV as the one they are testing the clear for. This makes sense for the purpose of the unit tests, I don't want to change them nor do I want to bloat our unit tests with a slight variantion of these 2 existing tests.

### DIFF
--- a/source/engine/framePass.cpp
+++ b/source/engine/framePass.cpp
@@ -341,34 +341,12 @@ HdTaskSharedPtrVector FramePass::GetRenderTasks(RenderBufferBindings const& inpu
     _lightingManager->SetLighting(_passParams.viewInfo.lights, _passParams.viewInfo.material,
         _passParams.viewInfo.ambient, _cameraDelegate.get(), _passParams.modelInfo.worldExtent);
 
-    // If clearBackgroundColor is false we assume we are rendering to an aov that is already initialized
-    // by a previous pass.  Skip the descriptor setup if that is the case.
-    if (_passParams.visualizeAOV == HdAovTokens->color)
-    {
-        if (_passParams.clearBackgroundColor)
-        {
-            _bufferManager->SetRenderOutputClearColor(
-                HdAovTokens->color, VtValue(_passParams.backgroundColor));
-        }
-        else
-        {
-            _bufferManager->SetRenderOutputClearColor(HdAovTokens->color, VtValue());
-        }
-    }
-
-    if (_passParams.visualizeAOV == HdAovTokens->depth)
-    {
-        if (_passParams.clearBackgroundDepth)
-        {
-            _bufferManager->SetRenderOutputClearColor(
-                HdAovTokens->depth, VtValue(_passParams.backgroundDepth));
-        }
-        else
-        {
-            _bufferManager->SetRenderOutputClearColor(HdAovTokens->depth, VtValue());
-        }
-    }
-
+    // Setup the clear parameters for color and depth. Empty VtValue() disables clearing the buffer.
+    _bufferManager->SetRenderOutputClearColor(HdAovTokens->color,
+        _passParams.clearBackgroundColor ? VtValue(_passParams.backgroundColor) : VtValue());
+    _bufferManager->SetRenderOutputClearColor(HdAovTokens->depth,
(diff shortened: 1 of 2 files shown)


sm-adsk:
We had an earlier issue that was fixed where the clear flag was "sticking" once it was enabled. Do these changes break that fix?

source/engine/framePass.cpp
```diff
@@ -341,34 +341,12 @@ HdTaskSharedPtrVector FramePass::GetRenderTasks(RenderBufferBindings const& inpu
...
-        {
-            _bufferManager->SetRenderOutputClearColor(HdAovTokens->depth, VtValue());
-        }
-    }
-
+    // Setup the clear parameters for color and depth. Empty VtValue() disables clearing the buffer.
+    _bufferManager->SetRenderOutputClearColor(HdAovTokens->color,
+        _passParams.clearBackgroundColor ? VtValue(_passParams.backgroundColor) : VtValue());
+    _bufferManager->SetRenderOutputClearColor(HdAovTokens->depth,
+        _passParams.clearBackgroundDepth ? VtValue(_passParams.backgroundDepth) : VtValue());
```
sebastienberube-adsk, line 348: This is \the same as before, except that clear values are set no matter what is currently visualized. I believe this is what we want, the currently visualized buffer should not influence how buffer are cleared, especially the depth buffer.

source/engine/renderBufferManager.cpp
```diff
@@ -144,9 +144,6 @@ class RenderBufferManager::Impl : public RenderBufferSettingsProvider
         HdRenderBuffer* depthBuffer, PXR_NS::HdRenderBuffer* neyeBuffer,
         const SdfPath& controllerId);
 
-    /// Resets the clear values
-    void ResetRenderOutputClear();
```
sebastienberube-adsk, line 148: No longer necessary, as we can now reset (disable) the render output clear operation by calling `RenderBufferManaer::SetRenderOutputClearColor()` with an empty `VtValue()`.

source/engine/renderBufferManager.cpp
```diff
@@ -251,9 +243,6 @@ bool RenderBufferManager::Impl::SetRenderOutputs(const TfTokenVector& outputs,
         }
     }
 
-    // Clear the AOV task cache to be able to change the clear background and depth values.
-    ResetRenderOutputClear();
```
sebastienberube-adsk, line 255: Controlling the render output clear color, including resetting them, is now done in the FramePass, outside of RenderBufferManager::Impl::SetRenderOutputs.

I believe this is simpler, less bug prone, and less redundant.
The call to ResetRenderOutputClear() here was also sometimes ignored, unless the list of input AOVs have been changed (the early out conditions at the top of the function were preventing the execution of this clear, so the call was sometimes inadvertently omitted).

rochaallanc (approved):
Looks good to me! 🚢 it!
Great arguments to change that.

PR #61, opened 2025-09-09
OGSMOD-8008: Conditional usage of Python debug lib (if found)

sebastienberube-adsk:
OGSMOD-8008 HVT - build issue with debug Python libs

**Context:**
The problem occurs when building in debug for developers who **have not** installed Python debug libraries, resulting in the following error: `cannot open file 'python[version]_d.lib'`
This error is caused by a code block in HVT [CMakeLists.txt#L239-L247](https://github.com/Autodesk/hydra-viewport-toolbox/blob/b4bb1196a3fce00ad28984f4eb0370b55f0631c3/CMakeLists.txt#L239-L247), which defines BOOST_DEBUG_PYTHON and forces linking with debug Python libraries:
```
add_compile_definitions(
$<$<CONFIG:Debug>:BOOST_DEBUG_PYTHON>
)
```

This code block is necessary for Windows users who have debug Python libraries installed, so it must be preserved (otherwise, the opposite problem can happen: `cannot open file 'python[version].lib'`). 

**The Fix:**
Added automatic detection of Python debug library availability using CMake's IMPORTED_LOCATION_DEBUG property. The BOOST_DEBUG_PYTHON definition is now only applied when debug libraries are actually present on the system, fixing both use cases and issues with and without Python debug libraries on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ if(NOT DEFINED OPENUSD_INSTALL_PATH AND DEFINED ENV{OPENUSD_INSTALL_PATH})
 endif()
 
 # Includes various helpers.
-include(Helpers)
+include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/Helpers.cmake")
 
 # Setups and runs vcpkg.
 include(VcpkgSetup)


CMakeLists.txt
```diff
@@ -236,15 +236,40 @@ if (MSVC)
...
-    # and will then include pyconfig.h :
-    # https://github.com/boostorg/python/blob/boost-1.81.0/include/boost/python/detail/wrap_python.hpp#L23-L57
-    # pyconfig.h will then autolink the Python lib; however, since _DEBUG was #undef'd, it will try to
-    # link against the release version of the Python lib, causing a linker error :
-    # https://github.com/python/cpython/blob/v3.11.4/PC/pyconfig.h#L269-L286
-    add_compile_definitions(
-        $<$<CONFIG:Debug>:BOOST_DEBUG_PYTHON>
-    )
+    # Function to detect if Python debug libraries are available.
+    function(detect_python_debug_libraries result_var)
```
DDoS, line 240: Move this to `Helpers.cmake` please.

CMakeLists.txt
```diff
@@ -236,15 +236,40 @@ if (MSVC)
...
+        # Check if the Python3::Python target has a debug location set.
+        get_target_property(python_debug_lib Python3::Python IMPORTED_LOCATION_DEBUG)
+            
+        if(python_debug_lib AND EXISTS "${python_debug_lib}")
+            set(${result_var} TRUE PARENT_SCOPE)
+        endif()
+    endfunction()
+
+    # Call function to check if Python debug libraries are available.
+    detect_python_debug_libraries(_PYTHON_DEBUG_AVAILABLE)
```
DDoS, line 259: Minor
```suggestion
    check_python_has_debug_library(_PYTHON_DEBUG_AVAILABLE)
```

CMakeLists.txt
```diff
@@ -236,15 +236,40 @@ if (MSVC)
...
-    # https://github.com/python/cpython/blob/v3.11.4/PC/pyconfig.h#L269-L286
-    add_compile_definitions(
-        $<$<CONFIG:Debug>:BOOST_DEBUG_PYTHON>
-    )
+    # Function to detect if Python debug libraries are available.
+    function(detect_python_debug_libraries result_var)
+        set(${result_var} FALSE PARENT_SCOPE)
+        
+        # Try to find Python3.
+        find_package(Python3 COMPONENTS Development QUIET)
```
DDoS, line 244: Since finding Python might have been done by the parent project, you should skip this if the Python target already exists (`if(TARGET <name>) ... endif()`).

DDoS: I guess you could also rely on `Python3_FOUND`.

debloip-adsk (commented):
Changes look good to me, I launched a MayaHydra build with an updated HVT with this fix to check if everything still works on our end.

CMakeLists.txt
```diff
@@ -236,15 +236,21 @@ if (MSVC)
...
+        # and will then include pyconfig.h :
+        # https://github.com/boostorg/python/blob/boost-1.81.0/include/boost/python/detail/wrap_python.hpp#L23-L57
+        # pyconfig.h will then autolink the Python lib; however, since _DEBUG was #undef'd, it will try to
+        # link against the release version of the Python lib, causing a linker error :
+        # https://github.com/python/cpython/blob/v3.11.4/PC/pyconfig.h#L269-L286
+        # The code below forces using Python debug libs, but should only be executed if Python debug libs
+        # were installed and are available.
+        add_compile_definitions(
+            $<$<CONFIG:Debug>:BOOST_DEBUG_PYTHON>
+        )
```
sebastienberube-adsk, line 252: @debloip-adsk @DDoS

Alternate way to handle this issue:

Given Boost was removed from USD, perhaps the parent project could handle setting this specific compile_definition instead of keeping this in HVT.

sebastienberube-adsk: I just had a look at the Boost dependency history in USD.

The external dependency on the full Boost library was indeed removed.
**However**, the some Boost code (boost::python) was copied to OpenUSD repo and modified by Pixar.

See, for example, this code from OpenUSD 25.08:
https://github.com/PixarAnimationStudios/OpenUSD/blob/v25.08/pxr/external/boost/python/detail/wrap_python.hpp#L29-L53

Long story short, we still have issues with wrap_python.hpp.

PR #62, opened 2025-09-09
Frameworks directory is sibling of executable in iOS Bundle, not in the parent directory

wingfiring:
The layout of iOS bundle is different from MacOS, and. the `Frameworks` folder is not in the parent folder of the executable.
This fix correct the resource path on iOS.

### DIFF
--- a/source/utils/pathUtils.cpp
+++ b/source/utils/pathUtils.cpp
@@ -136,9 +136,14 @@ std::filesystem::path GetDefaultMaterialXDirectory()
     // Root Location where MaterialX "libraries" are expected to be located.
     // For MacOS they are in Application Bundle/Contents/Frameworks and for others they are in the
     // same folder as the executable.
+    // For iOS, Frameworks is under the root of bundle, like "Application Bundle/Frameworks"
 #if defined(__APPLE__)
-    NSString* frameworkPath = [[[[NSBundle mainBundle] resourcePath]
-        stringByDeletingLastPathComponent] stringByAppendingPathComponent:@"Frameworks"];
+    NSBundle* bundle = [NSBundle mainBundle];
+    NSString* resourcePath = [bundle resourcePath];
+#if (TARGET_OS_IPHONE != 1)
+    resourcePath = [resourcePath stringByDeletingLastPathComponent];
+#endif
+    NSString* frameworkPath = [resourcePath stringByAppendingPathComponent:@"Frameworks"];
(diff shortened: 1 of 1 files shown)


source/utils/pathUtils.cpp
```diff
@@ -137,9 +137,18 @@ std::filesystem::path GetDefaultMaterialXDirectory()
     // For MacOS they are in Application Bundle/Contents/Frameworks and for others they are in the
     // same folder as the executable.
 #if defined(__APPLE__)
+#if (TARGET_OS_IPHONE == 1)
+    NSBundle* bundle = [NSBundle mainBundle];
+    NSString* resourcePath = [bundle resourcePath];
+    NSString* frameworkPath = [resourcePath stringByAppendingPathComponent:@"Frameworks"];
```
sebastienberube-adsk, line 143: [Observation]
Ok, so my understanding is this is an expansion of the expression below:
```
NSString* frameworkPath = [[[[NSBundle mainBundle] resourcePath]
        stringByDeletingLastPathComponent] stringByAppendingPathComponent:@"Frameworks"];
```        
where `stringByDeletingLastPathComponent` was removed.
In other words, the folder is a sibling, as described in your PR.

That makes sense.

sebastienberube-adsk: [Minor]

I give my approval without this, but perhaps the expression below (ln 148-149) chould also be expanded and written in the same style as yor change above. It would be slightly clearer IMHO.
i.e.

```
#if defined(__APPLE__)
    NSBundle* bundle = [NSBundle mainBundle];
    NSString* resourcePath = [bundle resourcePath];
#if (TARGET_OS_IPHONE == 1)
    NSString* frameworkPath = [resourcePath stringByAppendingPathComponent:@"Frameworks"];
#else
    NSString* parentPath = [resourcePath stringByDeletingLastPathComponent]
    NSString* frameworkPath = [parentPath stringByAppendingPathComponent:@"Frameworks"];
```

sebastienberube-adsk (approved):

hodoulp (approved):

hodoulp:
Note: Have the `HVTEx` and `Components` corresponding pull requests green before merging this one.

wingfiring:
> Note: Have the `HVTEx` and `Components` corresponding pull requests green before merging this one.↳

Verified in `HVTEx` and `Components`.

PR #65, opened 2025-09-12
OGSMOD-8012: Remove git-lfs from HVT

sm-adsk:
The goal of the pull request is to remove the `git lfs` support from the repository.
Steps were:
1. edit the `.gitattributes` to remove the lfs support
2. download all the formerly tracked files using the command `git lfs ls-files`


sm-adsk:
> Note: Can you try your branch in another clone to see what's up?

I tried it in a new clone and it's all good.

hodoulp (approved):

PR #66, opened 2025-09-16
OGSMOD-6304 - Fix the axis draw order by adding a small offset to the Y & Z axes.

sm-adsk:
The pull request changes the USD file of the tripod axes to avoid the z-depth fighting issue (which was adding instability to the unit tests). The Y and Z axes are now translated a little bit.


hodoulp:
There several aspects to consider with this improvement:
- On the three axes, ideally the depth of two axes must be slightly moved. For example the X axis is unchanged but Y and Z axes are.
- The problem randomly occurs, the challenge will then be to find a way to correctly validate the fix.

hodoulp:
⚠️ @sm-adsk How do you merge content from #68?

PR #69, opened 2025-09-18
Run the tests on Azure VM with GPU support

laforel:


.github/workflows/ci-test.yaml
```diff
@@ -0,0 +1,66 @@
...
+    
+  # Report status to PR
+  report-status:
+    needs: gpu-tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Report test results to PR
+        uses: actions/github-script@v7
+        with:
+          script: |
```
DDoS, line 30: There's no better way to do this than writing out a script manually?

.github/workflows/ci-test.yaml
```diff
@@ -0,0 +1,66 @@
...
+              repo: context.repo.repo,
+              sha: sha,
+              state: conclusion,
+              context: 'GPU Tests',
+              description: `GPU tests ${conclusion} for PR #${pr_number}`,
+              target_url: `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`
+            });
+            
+            // Comment on PR
+            await github.rest.issues.createComment({
```
DDoS, line 57: I don't think we need a comment, the PR status info is good enough, that's what it's for. Keep the comments space for reviewers.

source/CMakeLists.txt
```diff
@@ -104,6 +104,27 @@ if (APPLE OR LINUX)
     )
 endif()
 
+if (UNIX)
```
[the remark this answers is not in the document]
laforel: Yeah this one I wasn't sure. The hvt library was not finding usd dependencies, its RPATH was not right so I added this, thinking it might just be a problem in another cmake to expose something. The weird thing is it was working on most linux, but not on the new VM I spinned up in azure.

DDoS (approved):

PR #74, opened 2025-09-25
OGSMOD-8045: Add a way to override specific task update function, after creation.

sebastienberube-adsk:
Previously, the only way to set the task commit function was at creation, with the **CommitTaskFn** argument:
[TaskManager::AddTask(TfToken taskName, TParam initialParams, CommitTaskFn const& fnCommit)](https://github.com/Autodesk/hydra-viewport-toolbox/blob/d49868585e20ff6598b4c8f19d8d686f61a33078/include/hvt/engine/taskManager.h#L114-L117)

In some scenarios, we might want to update the commit function **AFTER** the task is created. A simple use case for allowing this is if some user wants to create default tasks with the function [hvt::CreateDefaultTasks](https://github.com/Autodesk/hydra-viewport-toolbox/blob/d49868585e20ff6598b4c8f19d8d686f61a33078/source/engine/taskCreationHelpers.cpp#L173-L272), but is unhappy with the way some commit function is implemented, then the ability to override a specific commit function is very handly.

This makes the TaskCreationHelper more flexible, and prevents having to reimplement most of [hvt::CreateDefaultTasks](https://github.com/Autodesk/hydra-viewport-toolbox/blob/d49868585e20ff6598b4c8f19d8d686f61a33078/source/engine/taskCreationHelpers.cpp#L173-L272) only to redefine a specific task commit function.


rochaallanc (approved):
Looks good to me!

These changes make sense to me since we should have flexibility to change the behavior, params in a later stage. Probably it will be useful mainly from an application point of view.

source/engine/taskManager.cpp
```diff
@@ -152,6 +152,18 @@ void TaskManager::EnableTask(TfToken const& instanceName, bool enable)
...
 
+void TaskManager::SetTaskCommitFn(TfToken const& taskName, CommitTaskFn const& fnCommit)
+{
+    TaskList::iterator it = GetTaskEntry(_tasks, taskName);
+    it->fnCommit          = fnCommit;
+}
+
+void TaskManager::SetTaskCommitFn(SdfPath const& uid, CommitTaskFn const& fnCommit)
+{
+    TaskList::iterator it = GetTaskEntry(_tasks, uid);
```
hodoulp, line 163: Manage the case when the task is not found.

hodoulp (changes requested):

hodoulp (approved):

PR #82, opened 2025-10-07
OGSMOD-8106 Fix unstable TestFramePasses unit tests

sebastienberube-adsk:
We recently realized some of our multi-pass unit tests were failing randomly:
- TestFramePasses_MultiViewports
- TestFramePasses_ClearColorBuffer
- TestFramePasses_TestDynamicAovInputs

This tends to happen more often when 2 consecutive frame passes that have a different lighting and camera settings, and our assumption was a buffer sync issue between the 1st and the 2 frame pass.

Waiting for the GPU to be idle before starting the 2nd frame pass is quite blunt, and we will have to implement a more tailored sync mechanism for supporting consecutive frame passes.

This is a transient solution.

### DIFF
--- a/test/RenderingFramework/AndroidTestContext.cpp
+++ b/test/RenderingFramework/AndroidTestContext.cpp
@@ -88,6 +88,11 @@ void VulkanRendererContext::run(
     }
 }
 
+void VulkanRendererContext::waitForGPUIdle()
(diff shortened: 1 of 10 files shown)


AdamFelt (approved):
LGTM

PR #83, opened 2025-10-09
OGSMOD-8090: Enable Vulkan unit tests in Viewport Toolbox

sm-adsk:
### Description:
This PR enables Vulkan unit tests in the Hydra Viewport Toolbox by addressing compatibility issues and updating test infrastructure to support Vulkan alongside the existing OpenGL backend.

### Key changes:
- Fixed multi-frame pass handling for Vulkan rendering
- Updated baseline images for various tests across platforms (macOS, iOS, Android)
- Improved blur shader implementation and FXAA test stability
- Enhanced test context synchronization (QueueWaitIdle → DeviceWaitIdle)
- Adjusted pixel count thresholds for platform-specific test variations
- Fixed frame framing behavior in multipass and multi-viewport tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![CI Full: (Mac-Linux-Windows Deb-Rel)](https://github.com/Autodesk/hydra-viewport-toolbox/actions/workflows/ci-full.yaml/badge.svg?branch=main)](https://github.com/Autodesk/hydra-viewport-toolbox/actions/workflows/ci-full.yaml?query=branch%3Amain+workflow%3A%22CI%20Full%22)
+[![CI GPU Tests: (Linux Rel)](https://github.com/Autodesk/hydra-viewport-toolbox/actions/workflows/ci-test.yaml/badge.svg?branch=main)](https://github.com/Autodesk/hydra-viewport-toolbox/actions/workflows/ci-test.yaml?query=branch%3Amain+workflow%3A%22CI%20GPU%20Tests%22)
 # Hydra Viewport Toolbox
 The **Hydra Viewport Toolbox** (HVT) is a library of utilities that can be used by an application to simplify the use of [OpenUSD](https://openusd.org) Hydra for the application's graphics viewports. The utilities can be used together or independently to add common viewport functionality and improve the performance and visual quality of viewports.
 


test/RenderingFramework/CMakeLists.txt
```diff
@@ -94,6 +102,22 @@ target_link_libraries(${_TARGET}
         ${glew_dep}
 )
 
+# Link with SDL2 only if Vulkan is enabled since SDL2 is required by the tests.
+if(HVT_ENABLE_VULKAN)
+    target_link_libraries(${_TARGET}
+        PUBLIC
+            $<TARGET_NAME_IF_EXISTS:SDL2::SDL2main>
```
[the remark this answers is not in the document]
sm-adsk: The SDL2 package prints this info when using CMake configure so I used it as is:
```
sdl2 provides CMake targets:

    find_package(SDL2 CONFIG REQUIRED)
    target_link_libraries(main
        PRIVATE
        $<TARGET_NAME_IF_EXISTS:SDL2::SDL2main>
        $<IF:$<TARGET_EXISTS:SDL2::SDL2>,SDL2::SDL2,SDL2::SDL2-static>
    )
```
But I have simplified it.

hodoulp (changes requested):
Note: I need time to review this work.

test/RenderingFramework/CMakeLists.txt
```diff
@@ -79,21 +87,30 @@ target_compile_definitions(${_TARGET}
...
+# Link with SDL2 only if Vulkan is enabled since SDL2 is required by the tests.
+if (HVT_ENABLE_VULKAN)
+    
+    target_link_libraries(${_TARGET}
+        PUBLIC
+            $<IF:$<TARGET_EXISTS:SDL2::SDL2>,SDL2::SDL2,SDL2::SDL2-static>
+    )
+    
+    # Add VMA include directory to fix vk_mem_alloc.h include issue.
+    find_package(VulkanMemoryAllocator CONFIG REQUIRED)
```
DDoS: `find_package` and similar calls should be grouped in the same file to prevent hiding dependencies throughout the CMake script.

test/RenderingFramework/TestHelpers.cpp
```diff
@@ -62,7 +62,7 @@ const std::filesystem::path inAssetsPath = getenv("HVT_TEST_ASSETS");
 const std::string resFullpath            = getenv("HVT_RESOURCES");
 std::filesystem::path inBaselinePath     = getenv("HVT_BASELINES");
 #else
-const std::filesystem::path outFullpath  = TOSTRING(TEST_DATA_OUTPUT_PATH) + "/computed";
+const std::filesystem::path outFullpath  = TOSTRING(TEST_DATA_OUTPUT_PATH) + "computed";
```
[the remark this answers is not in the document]
DDoS:
```cpp
const auto outFullpath  = std::filesystem::path(TOSTRING(TEST_DATA_OUTPUT_PATH)) / "computed";
```
Should work and be more clear.

test/RenderingFramework/VulkanTestContext.cpp
```diff
@@ -896,7 +899,7 @@ void VulkanTestContext::init()
 {
     namespace fs = std::filesystem;
 
-    _sceneFilepath = TOSTRING(TEST_HVT_DATA_PATH) + "/data/assets/usd/test_fixed.usda";
+    _sceneFilepath = TOSTRING(HVT_TEST_DATA_PATH) + "/data/assets/usd/test_fixed.usda";
```
DDoS:
```suggestion
    _sceneFilepath = fs::path(TOSTRING(HVT_TEST_DATA_PATH)) / "data/assets/usd/test_fixed.usda";
```

test/tests/testFramePass.cpp
```diff
@@ -224,6 +224,11 @@ void TestDynamicFramePassParams(
 
         sceneFramePass->Render();
 
+        // Force GPU sync.
+        // Note: It greatly improves the image result consistency on some backends which is critical
```
DDoS: Same comment as before. There are many other such cases, so to avoid too many comments I won't mention them any further, but the comment still applies.

test/tests/testSearches.cpp
```diff
@@ -245,7 +245,15 @@ HVT_TEST(TestViewportToolbox, TestSearchFaces)
     computedFileName = "origin_dev/02505/" + computedFileName;
 #endif
 
-    ASSERT_TRUE(context->validateImages(computedImageName, TestHelpers::gTestNames.fixtureName));
+    uint8_t threshold           = 1;
+    uint8_t pixelCountThreshold = 1;
```
DDoS: `uint8_t` doesn't sound like the right type.

test/CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ find_package(GTest CONFIG REQUIRED)
 find_package(glfw3 CONFIG REQUIRED)
 find_package(glew CONFIG REQUIRED)
 
+# Only if Vulkan is enabled since SDL2 is required by the tests.
+if(HVT_ENABLE_VULKAN)
+    find_package(SDL2 CONFIG REQUIRED)
```
DDoS: Group this with other `find_package` calls.

CMakePresets.json
```diff
@@ -35,6 +35,26 @@
         "CMAKE_BUILD_TYPE": "Release"
       }
     },
+    {
+      "name": "debug-vulkan",
```
hodoulp: To be consistent with other preset it should be `debugwithvulkan`

test/tests/testSearches.cpp
```diff
@@ -311,29 +319,47 @@ HVT_TEST(TestViewportToolbox, TestSearchEdges)
     ASSERT_EQ(primState->pointIndices.size(), 0);
     ASSERT_EQ(primState->pointColorIndices.size(), 0);
 
-    static const std::vector<VtIntArray> results
+    std::vector<VtIntArray> results;
 #if defined(_WIN32) || defined(__linux__)
```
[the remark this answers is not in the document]
hodoulp: You are right. But macOS & iOS are both metal only (i.e., no need to check for the backend).
And the Linux case is OpenGL only.

test/tests/testSearches.cpp
```diff
@@ -245,7 +245,15 @@ HVT_TEST(TestViewportToolbox, TestSearchFaces)
     computedFileName = "origin_dev/02505/" + computedFileName;
 #endif
 
-    ASSERT_TRUE(context->validateImages(computedImageName, TestHelpers::gTestNames.fixtureName));
+    uint8_t threshold           = 1;
+    uint8_t pixelCountThreshold = 1;
+#if defined(_WIN32) && defined(ENABLE_VULKAN)
```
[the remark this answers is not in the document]
hodoulp: The line `GetParam() == TestHelpers::RenderingBackend::Vulkan` is a test on the backend used for the test and only the Windows unit test framework supports `Vulkan` for now.

test/RenderingFramework/TestFlags.h
```diff
@@ -11,36 +11,87 @@
...
+        const std::string computedImageName = TestHelpers::appendParamToImageFile(imageFile);     \
+        HVTTest##TestName(computedImageName, imageFile);                                          \
+    }                                                                                             \
+    void TestName::HVTTest##TestName(                                                             \
+        [[maybe_unused]] const std::string& computedImageName,                                    \
         [[maybe_unused]] const std::string& imageFile)
+#else
+
+    #define GetParam() ([]() -> TestHelpers::RenderingBackend {                                         \
+        return static_cast<TestHelpers::RenderingBackend>(-1); }())
```
hodoulp: This is problematic it should be the default backend for each platform.
Windows-> OpenGL
MacOS, iOS -> Metal
etc.

1) A proposal could be:
```
enum class RenderingBackend
{
#ifdef _WIN32 || __linux__
    OpenGL
#elif __APPLE__
    Metal
#else
    OpenGL
#endif
    Vulkan,
};

inline std::string renderingBackendToString(RenderingBackend backend)
    {
        switch (backend)
        {
            case RenderingBackend::Vulkan:
                return "Vulkan";
            case RenderingBackend::OpenGL:
                return "OpenGL";
            default:
                return "Unknown";
        }
    }
```

2) There is already tokens for all the backends. Why having these two helpers when it already exists?

rochaallanc: This suggestion makes sense.

test/tests/testSearches.cpp
rochaallanc: 👍

test/tests/testSearches.cpp
```diff
@@ -245,7 +245,15 @@ HVT_TEST(TestViewportToolbox, TestSearchFaces)
...
+    if (GetParam() == TestHelpers::RenderingBackend::Vulkan) {
+        pixelCountThreshold = 99;
```
rochaallanc: 99? is the number arbitrary?

rochaallanc (commented):
Overall looks good to me, great job! 
Just added some few minor comments. There are some comments from others I think are relevant.

test/tests/testSearches.cpp
sm-adsk: That's the exact number of pixels that were different.

test/tests/testSearches.cpp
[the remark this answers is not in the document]
DDoS: Avoid hardcoding assumptions. If you hardcode Metal and macOS and iOS you'll break Vulkan support, which is very convenient for development and testing, even if we don't plan to officially support it.

A while ago I spent quite a bit of time disentangling the web of runtime and compile-time checks to make this project more flexible, let's not go back down that path.

test/tests/testFramePass.cpp
```diff
@@ -405,33 +417,33 @@ TEST(TestViewportToolbox, TestFramePassSelectionSettingsProvider)
...
+        auto& framePassParams = framePass->params();
 
-    framePassParams.enableSelection       = false;
-    framePassParams.enableOutline         = false;
-    framePassParams.selectionColor        = GfVec4f(1.0f, 0.0f, 0.0f, 1.0f); // Red
-    framePassParams.locateColor           = GfVec4f(0.0f, 1.0f, 0.0f, 1.0f); // Green
+        framePassParams.enableSelection       = false;
+        framePassParams.enableOutline         = false;
+        framePassParams.selectionColor        = GfVec4f(1.0f, 0.0f, 0.0f, 1.0f); // Red
+        framePassParams.locateColor           = GfVec4f(0.0f, 1.0f, 0.0f, 1.0f); // Green
```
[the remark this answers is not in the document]
sm-adsk: Done.

include/hvt/resources/shaders/blur.glslfx
```diff
@@ -66,7 +64,7 @@ void main(void)
         for(uint i=1u; i<=Quality; i++)
         {
             float coeff   = i*inv_quality;
-            vec2 uvOffset = (vec2(cos(d),sin(d))*Radius*coeff) * screenSize;
+            vec2 uvOffset = vec2(cos(d), sin(d)) * Size * coeff;
```
hodoulp: Good catch

CMakeLists.txt
```diff
@@ -325,6 +325,10 @@ if(HVT_ENABLE_VULKAN)
     add_compile_definitions(ENABLE_VULKAN)
     find_package(VulkanLoader CONFIG REQUIRED)
     find_package(VulkanHeaders CONFIG REQUIRED)
+    # Add VMA include directory to fix vk_mem_alloc.h include issue.
+    find_package(VulkanMemoryAllocator CONFIG REQUIRED)
+    # SDL2 is required by the Vulkan tests.
+    find_package(SDL2 CONFIG REQUIRED)
```
DDoS: Missing `ENABLE_TESTS` check.

PR #85, opened 2025-10-14
AGPHVT-82 : Improve buffer transitions and content copying between frame passes.

AdamFelt:
This submission fixes render buffer handling between renderers and frame passes.

Removes the CopyTask and moves copy functionality to the RenderBufferManager.
Moves AOVInput and RenderTask gathering out of the update function and puts it in the render loop.
Texture output contents are now copied to the next pass buffer inputs between passes, not as part of the pass.
RenderBuffers are reused from the previous pass not the first pass which supports more flexible mixing of renderers.
New render buffers are created if the render switches between passes so renderers are no longer rendering into a render buffer they did not create. RenderBuffers are still reused from previous passes if the renderer does not change.


source/engine/framePass.cpp
```diff
@@ -246,20 +246,32 @@ std::tuple<SdfPathVector, SdfPathVector> FramePass::CreatePresetTasks(PresetTask
 
     if (!IsStormRenderDelegate(GetRenderIndex()) && _bufferManager->IsAovSupported())
     {
-        // Initialize the AOV system to render color, depth and ID buffers.
-        _bufferManager->SetRenderOutputs(
-            { HdAovTokens->color, HdAovTokens->depth, HdAovTokens->primId, HdAovTokens->elementId,
-                HdAovTokens->instanceId },
-            {},
-            _passParams.renderParams.viewport); // NOTE: this is the non-adjusted viewport.
-
```
sebastienberube-adsk: I never understood the exact reason why initializing AOV buffers was necessary when using a renderer other than Storm.

I can see similar code being executed with the TaskController, from the constructor:
`HdxTaskController::HdxTaskController` -> `HdxTaskController::_CreateRenderGraph()` -> `HdxTaskController::SetRenderOutputs("color")`

Since I do not understand why this line had been added to the TaskController, I am slightly worried removing it here might break something.

**However**, on the other hand, I can see that `_bufferManager->SetRenderOutputs` is called early, in the beggining of `FramePass::GetRenderTasks`. Furthermore, all task update code (the commit functions) are called at the very end of `FramePass::GetRenderTasks`, which means **not much is happening before** `SetRenderOutputs` is called. So maybe it does not change anything to make this call later, given the order of things with the FramePass.

In conclusion, removing `_bufferManager->SetRenderOutputs` from this task creation function in the FramePass is probably safe, and this call might only be needed with the TaskController because it has a different task parameter update sequence, requiring the early call to `HdxTaskController::SetRenderOutputs("color")`.

See:
https://git.autodesk.com/autodesk-forks/USD/blob/d34c093b8126e5b82fa35b478586539f82ceadd7/pxr/imaging/hdx/taskController.cpp#L292C13-L292C29

source/engine/renderBufferManager.cpp
```diff
@@ -252,57 +447,61 @@ bool RenderBufferManager::Impl::SetRenderOutputs(const TfTokenVector& outputs,
...
+    // Temporary 2D dimensions to calculate dimensions3 (the 2D version isn't used later).
+    const GfVec2i dimensions = _renderBufferSize != GfVec2i(0)
+        ? _renderBufferSize
+        : GfVec2i(static_cast<int>(viewport[2]), static_cast<int>(viewport[3]));
+    const GfVec3i dimensions3(dimensions[0], dimensions[1], 1);
+
+    HdAovDescriptorList outputDescs;
 
     // NOTE: A function could be used to get localOutputs.
     TfTokenVector localOutputs = outputs;
```
sebastienberube-adsk: [Observation]

This is out of scope for this PR since the name did not change, but we could rename `localOutputs` to something more descriptive of what it actually is, something like `supportedOutputs` maybe, since the only difference that I see between `outputs` and `localOutputs` is that each entry is validated with as shown below:

```
for (auto it = localOutputs.begin(); it != localOutputs.end();)
{
    HdAovDescriptor desc = _pRenderIndex->GetRenderDelegate()->GetDefaultAovDescriptor(*it);
    if (desc.format == HdFormatInvalid)
    {
        // The backend doesn't support this AOV, so skip it.
        it = localOutputs.erase(it);
    }
```

source/engine/renderBufferManager.cpp
```diff
@@ -252,57 +447,61 @@ bool RenderBufferManager::Impl::SetRenderOutputs(const TfTokenVector& outputs,
...
             std::equal(inputs.begin(), inputs.end(), _aovInputs.begin(), _aovInputs.end()))
         {
-            return false;
+            somethingChanged = false;
         }
     }
 
-    // If progressive rendering is enabled, render buffer clear is only required when `_aovOutputs
-    // != outputs`.
-    bool needClear = !_isProgressiveRenderingEnabled || _aovOutputs != outputs;
```
sebastienberube-adsk: [Observation]

`bool needClear = !_isProgressiveRenderingEnabled || _aovOutputs != outputs;`

I see this was moved below and is only executed `if (somethingChanged)` which should be functionally the same as before, with the early return if cached inputs and outputs have changed.

source/engine/renderBufferManager.cpp
```diff
@@ -252,57 +447,61 @@ bool RenderBufferManager::Impl::SetRenderOutputs(const TfTokenVector& outputs,
...
-
     _aovOutputs = outputs;
 
-     _aovInputs.clear();
-    if (inputs.size() > 0)
-    {
-        std::copy(inputs.begin(), inputs.end(), back_inserter(_aovInputs));
-        _viewportAov = TfToken();
-    }
```
sebastienberube-adsk: I see this block was also moved below and is only executed `if (somethingChanged)`, which makes sense and should deliver the same outcome as before:

```
     _aovInputs.clear();
    if (inputs.size() > 0)
    {
        std::copy(inputs.begin(), inputs.end(), back_inserter(_aovInputs));
        _viewportAov = TfToken();
    }

```

source/engine/renderBufferManager.cpp
```diff
@@ -252,57 +447,61 @@ bool RenderBufferManager::Impl::SetRenderOutputs(const TfTokenVector& outputs,
...
+    // Temporary 2D dimensions to calculate dimensions3 (the 2D version isn't used later).
+    const GfVec2i dimensions = _renderBufferSize != GfVec2i(0)
+        ? _renderBufferSize
+        : GfVec2i(static_cast<int>(viewport[2]), static_cast<int>(viewport[3]));
+    const GfVec3i dimensions3(dimensions[0], dimensions[1], 1);
```
sebastienberube-adsk: [Minor/Cosmetic]

I see this code to initialized `dimensions3` is the same as before, but it was moved up in the function.
It is only used ~100 lines later.

This is **extra minor** but it would make this PR a tiny bit easier to read if it was kept closer to where it is used, a bit lower.

In the end, it has no actual impact wherever is it initialized, I am mostly adding this comment as a guide comment/reference when consulting this PR.

source/engine/renderBufferManager.cpp
```diff
@@ -321,25 +520,58 @@ bool RenderBufferManager::Impl::SetRenderOutputs(const TfTokenVector& outputs,
...
+                        // FUTURE: We may want to revisit this decision in the future.  
+                        // The long-term solution may be to do post processing at the sub-pixel accuracy.
+                        depthInput.texture = HgiTextureHandle();
+                    }
+                }
+                else if (!colorInput.texture)
+                {
+                    colorDesc  = desc;
+                    colorInput = input;
+                }
```
sebastienberube-adsk: I understand that "color" is usually listed first in the inputs vector, but would it make sense to validate this assumption by checking if or asserting that `localOutputs[i] == pxr::HdAovTokens->color` ?

source/engine/renderBufferManager.cpp
```diff
@@ -362,20 +598,27 @@ bool RenderBufferManager::Impl::SetRenderOutputs(const TfTokenVector& outputs,
...
+
+        aovBindingsClear[i].aovName    = localOutputs[i];
+        aovBindingsClear[i].clearValue = !foundInput.buffer ? outputDescs[i].clearValue : VtValue();
         aovBindingsClear[i].renderBufferId = GetAovPath(controllerId, localOutputs[i]);
         aovBindingsClear[i].aovSettings    = outputDescs[i].aovSettings;
-        aovBindingsClear[i].renderBuffer   = foundInput;
+
+        // Note, it would be better to just assign the output buffer here, but this breaks some
+        // unit tests that expect this to be null and do a pointer-as-string comparison if it is not 
+        // which is not easily fixable.
```
sebastienberube-adsk: [Question]

Which test is this breaking?
(If the test is `TestTaskParameters`, I could probaly do something about it)

In any case, this appears like a minor thing, won't block the PR for this, I am mostly curious of the test this would break.

source/engine/framePass.cpp
```diff
@@ -246,20 +246,32 @@ std::tuple<SdfPathVector, SdfPathVector> FramePass::CreatePresetTasks(PresetTask
...
         _selectionHelper->SetVisualizeAOV(_bufferManager->GetViewportAov());
     }
 
     return { taskIds, renderTaskIds };
 }
 
+hvt::RenderBufferBindings FramePass::GetRenderBufferBindingsForNextPass(
+    std::vector<pxr::TfToken> const aovs, bool copyContents)
+{
+    std::string renderName = GetRenderIndex()->GetRenderDelegate()->GetRendererDisplayName();
```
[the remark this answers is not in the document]
sebastienberube-adsk: Actually, a TfToken would be preferable here I think. The TfToken for the rendererName could be stored in the FramePass of the RenderBufferManager in FramePass::Initialize.

source/engine/renderBufferManager.cpp
AdamFelt: Yeah, we just can't bail out early because we need to do the copy, but the functionality should still be skipped.

source/engine/renderBufferManager.cpp
```diff
@@ -386,40 +629,28 @@ bool RenderBufferManager::Impl::SetRenderOutputs(const TfTokenVector& outputs,
...
-        for (auto input : inputs)
-        {
-            // This is super fragile and limited.
-            // We should be able to pass more inputs to other passes.
-            if (input.first == localOutputs[0])
-                firstInput = input.second;
-            if (input.first == "depth")
-                depthInput = input.second;
-            if (input.first == HdAovTokens->Neye)
-                neyeInput = input.second;
```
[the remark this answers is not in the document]
AdamFelt: Yeah, this is no longer needed.

source/tasks/aovInputTask.cpp
```diff
@@ -147,15 +147,12 @@ void AovInputTask::Execute(HdTaskContext* ctx)
...
 
-    static const TfToken msaaToken("colorMSAA");
-
     // Start by clearing aov texture handles from task context.
     // These are last frames textures and we may be visualizing different aovs.
     ctx->erase(HdAovTokens->color);
     ctx->erase(HdAovTokens->depth);
     ctx->erase(HdxAovTokens->colorIntermediate);
     ctx->erase(HdAovTokens->Neye);
-    ctx->erase(msaaToken);
```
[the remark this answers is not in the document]
AdamFelt: Yeah, I didn't touch Neye.  It still needs to be fixed.  It should not be on the context.  The context is for things that are read and rewritten, not just read.

PR #101, opened 2025-11-10
OGSMOD-4820 - Fix inconsistency for SSAO utest

hodoulp:
The pull request reenables `SSAO` unit test on macOS to investigate/fix instabilities.


kohakukun (commented):
The new image for FXAA looks kinda broken
image


Is this expected?

kohakukun:
Still looks a little bit weird but, if the author thinks that is the expected value we could circle back at some point to improve the settings of the task

PR #139, opened 2026-02-06
Update CMakeLists.txt

hamedsabri:
## Description
The @_MODULE_NAME@ isn't populated in the hvtConfig.cmake causing problem with consuming hvt in another project.


hodoulp (commented):
Good catch

hodoulp:
Will test on macOS, Windows and merge it.

PR #153, opened 2026-03-12
AGP-231 - Migrate TaskManager to scene index

hodoulp:
## Description

The pull request radically changes the `TaskManager` to work with `HdRetainedSceneIndex` only.

### Changes Made

- Refactors the `TaskManager` to uses `pxr::HdRetainedSceneIndex` (instead of the `pxr::HdSceneDelegate`) to store rendering data & to propagate dirtiness.  
- Refactors the `RenderBufferManager` and `LightManager` to now use the `pxr::HdRetainedSceneIndex` instance.
- Refactors the `FramePass` to manage the `pxr::HdRetainedSceneIndex` instance. 
- Adds `dosc/scene-index-task-manager.md` explaining the design change.
- Changes some unit test output filenames (i.e., prepended with a keyword) so it's easier find them in the computed directory. 
- Disables some output messages for debugging purpose from a previous pull request.

### Ease the review

A way to only review pertinent changes:
1. `dosc/taskmgr.md`
2. `TaskManager`
3. `RenderBufferManager`, `LightManager`

## Tests Performed

- [X] Tested with different render delegates

## Documentation

## Checklist


include/hvt/engine/taskManager.h
```diff
@@ -173,7 +174,8 @@ class HVT_API TaskManager
     PXR_NS::VtValue GetTaskValue(PXR_NS::SdfPath const& uid, PXR_NS::TfToken const& key);
 
     /// Sets the task value with the specified task unique identifier and key.
-    void SetTaskValue(
+    /// \return True if the value was accepted, false on error (empty uid, missing task, bad key).
+    [[nodiscard]] bool SetTaskValue(
```
sebastienberube-adsk: I am not sure the `[[nodiscard]]` is a good idea here:
- It can make sense to call this function without using the return value. It is up the user to decide.
- Updating HVT will create new warnings in existing applications that were calling the previous version that was returning `void`.

sebastienberube-adsk (approved):
Nice change! I am happy to see the transition to the Scene Index did not change the FramePass, TaskManager and TaskCreationHelper external facing interface too much. This is good news.

Overall, the changes make sense. The SceneIndex is now used instead of the SceneDelegate as a container for BufferDescriptors, Tasks, LightPrims. The bulk of the existing logic is the same, with some adaptations to deal with DataSouces (value wrapper) and the SceneIndex.

I did a resonably thorough review of all files, except for the LightingManager. There are too many changes there, I can't make much sense of it. I see some similarity in the new code with OpenUSD's taskControllerSceneIndex (e.g. [_LightSchemaDataSource](https://git.autodesk.com/autodesk-forks/USD/blob/21768172ecb79ec0e581b8ac310eda50e99d7f03/pxr/imaging/hdx/taskControllerSceneIndex.cpp#L744-L1058)) but the mapping is not 1:1. I am guessing the new LightingManager is a mashup of our previous restructuring and the refactor that happened on OpenUSD's side.

Finally, perhaps two minor global comments (nitpick):
- A lot of comments were removed, sometimes for no reason. I don't mind all that much, but this is just an observation. Was this intentional?
- You can keep the current PR that way, but there are also a lot of formatting and white spaces changes mixed up with this big refactor (e.g. taskCreationHelpers). It makes the review more difficult / straining and it is easier to review when they are separated. But you don't need to change it now, at least not for me, my review is done and the final result looks good +1

PR #159, opened 2026-04-01
Lighting regression

AdamFelt:
### Problem
Loading any scene caused significantly degrading performance. 

### Root Cause
Two issues in `lightingManager.cpp`:
1. **`ReplaceLightSprim` used Remove + Add for every update.** This sent `PrimsRemoved` followed by `PrimsAdded`.  This can be interpreted by a renderer as a light count change which can be invasive.  This happened even for simple parameter updates where the light already existed.
2. **Camera-light transforms were re-pushed every frame unconditionally.** The camera-light update path called `ReplaceLightSprim` every frame when `viewInvMatrix != identity`, without checking whether the composed transform actually changed.
### Fix
- **`ReplaceLightSprim`**: Removed the `RemoveLightSprim` call. `HdRetainedSceneIndex::AddPrims` on an existing path updates the data source in place and sends `PrimsAdded`, which observers handle as an update (not a structural add/remove cycle).
- **Camera-light guard**: Added a `_lastCameraLightTransforms` cache to store the last composed `viewInvMatrix * lightTransform` per light path. `ReplaceLightSprim` is now only called when the transform actually changes.
- **`RemoveLightSprim`**: Cleans up the new cache on light removal.
### Files Changed
- `source/engine/lightingManager.cpp`

### DIFF
--- a/source/engine/lightingManager.cpp
+++ b/source/engine/lightingManager.cpp
@@ -394,6 +394,9 @@ class LightingManager::Impl
     std::unordered_map<SdfPath, HdxShadowMatrixComputationSharedPtr, SdfPath::Hash>
         _shadowMatrixComputations;
 
+    // Last composed camera-light transform per path for change detection.
+    std::unordered_map<SdfPath, GfMatrix4d, SdfPath::Hash> _lastCameraLightTransforms;
+
 public:
     explicit Impl(SdfPath const& lightRootPath, HdRenderIndex* pRenderIndex,
         HdRetainedSceneIndexRefPtr const& retainedSceneIndex, bool isHighQualityRenderer) :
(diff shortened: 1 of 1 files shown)


source/engine/lightingManager.cpp
```diff
@@ -471,15 +474,20 @@ void LightingManager::Impl::RemoveLightSprim(size_t pathIdx)
         SdfPath const& path = _lightIds[pathIdx];
         _retainedSceneIndex->RemovePrims({ { path } });
         _lightData.erase(path);
+        _lastCameraLightTransforms.erase(path);
         _shadowMatrixComputations.erase(path);
     }
 }
 
-void LightingManager::Impl::ReplaceLightSprim(size_t pathIdx, GlfSimpleLight const& light,
+void LightingManager::Impl::ReplaceLightSprim(size_t /*pathIdx*/, GlfSimpleLight const& light,
```
sebastienberube-adsk, line 482: [Minor]

Why not remove the `pathIdx` parameter, if it is not used.
The function is private, it will not impact any external user.

sebastienberube-adsk (approved):
Approved

I was wondering how hacky it is to treat PrimsAdded as an update, but it turns out this is a well-known idiom in Hydra. Hydra observers are expected to treat PrimsAdded on a known path as "this prim's data has been replaced, re-pull it."

PR #160, opened 2026-04-01
AGP-279 - Light - Fix performance regression

hodoulp:
## Description

The pull request fixes a performance regression for the light management introduced in the pull request #153.

The design flaw is around the light management where any light updates were in fact, deleting and re-creating lights even if nothing change. That approach is only fine when the light type changed but in any other cases, that's only a light update.

Tested in Flash by @AdamFelt 

## Tests Performed
- [X] Tested with different render delegates

## Documentation

## Checklist


source/engine/lightingManager.cpp
```diff
@@ -501,10 +499,58 @@ void LightingManager::Impl::ReplaceLightSprim(size_t pathIdx, GlfSimpleLight con
         shadowParams->shadowMatrix          = shadowMatrixComp;
         _shadowMatrixComputations[pathName] = shadowMatrixComp;
     }
+    else
+    {
+        _shadowMatrixComputations.erase(pathName);
+    }
```
hodoulp, line 506: Missing cleanup.

source/engine/lightingManager.cpp
```diff
@@ -689,18 +735,19 @@ void LightingManager::SetLighting(GlfSimpleLightVector const& lights,
...
         lightingState->SetUseLighting(true);
         lightingState->SetLights(lights);
         lightingState->SetSceneAmbient(ambient);
         lightingState->SetMaterial(material);
     }
     else
     {
-        _impl->GetLightingContext()->SetUseLighting(false);
+        lightingState->SetUseLighting(false);
+        lightingState->SetLights({});
```
hodoulp, line 755: Do we need to also reset (to some default values) the `ambient` and `material`?

AdamFelt: Ambient and Material don't change when you change the light set.  I think this is fine as-is.

source/engine/lightingManager.cpp
```diff
@@ -501,10 +500,62 @@ void LightingManager::Impl::ReplaceLightSprim(size_t pathIdx, GlfSimpleLight con
...
     }
+    else
+    {
+        _shadowMatrixComputations.erase(pathName);
+    }
 
     auto lightPtr = std::make_shared<GlfSimpleLight>(light);
     std::shared_ptr<HdxShadowParams const> shadowParamsConst = shadowParams;
 
+    // Update the existing prim in place when possible.
```
AdamFelt, line 511: I guess my concern with making things more granular s is making sure we captured everything.  For instance you are setting shadow params, but not setting a locator for that.  Is this a problem, I don't know.

AdamFelt (approved):
I tested this and it does work for my use case.

PR #162, opened 2026-04-22
AGP-338 - Add signing

hodoulp:
## Description

The pull request adds the missing code signing on macOS.

### Tests Performed

## Documentation

## Checklist

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -116,6 +116,27 @@ endif()
 
 #---------------------------------------------------------------------------------------------------
 
+# macOS codesign
+if (APPLE)
+    if (DEFINED codeSignID)
+        set(HVT_CODESIGN_IDENTITY "${codeSignID}" CACHE STRING "Codesign identity for ${_TARGET} dylib on macOS")
+    else()
+        set(HVT_CODESIGN_IDENTITY "-" CACHE STRING "Codesign identity for ${_TARGET} dylib on macOS")
+    endif()
+
+    # Post-build codesign of the output dylib so dyld treats it as signed
+    if (HDFLASH_CODESIGN)
+        add_custom_command(TARGET ${_TARGET} POST_BUILD
+            COMMAND /usr/bin/codesign --force --sign "${HVT_CODESIGN_IDENTITY}" --timestamp=none
+                    "$<TARGET_FILE:${_TARGET}>"
+            COMMENT "Codesigning ${_TARGET} dylib ($<TARGET_FILE:${_TARGET}>)"
+            VERBATIM
+        )
+    endif()
+endif()
+
+#---------------------------------------------------------------------------------------------------
+
 # Prepare for creating relocatable packages.
 include(CMakePackageConfigHelpers)
 
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -122,16 +122,15 @@ endif()
 
 # macOS codesign
 if (APPLE)
-    if (DEFINED codeSignID)
-        set(HVT_CODESIGN_IDENTITY "${codeSignID}" CACHE STRING "Codesign identity for ${_TARGET} dylib on macOS")
-    else()
-        set(HVT_CODESIGN_IDENTITY "-" CACHE STRING "Codesign identity for ${_TARGET} dylib on macOS")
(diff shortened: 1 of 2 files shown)


source/CMakeLists.txt
```diff
@@ -116,6 +116,27 @@ endif()
...
+# macOS codesign
+if (APPLE)
+    if (DEFINED codeSignID)
+        set(HVT_CODESIGN_IDENTITY "${codeSignID}" CACHE STRING "Codesign identity for ${_TARGET} dylib on macOS")
+    else()
+        set(HVT_CODESIGN_IDENTITY "-" CACHE STRING "Codesign identity for ${_TARGET} dylib on macOS")
+    endif()
+
+    # Post-build codesign of the output dylib so dyld treats it as signed
+    if (HDFLASH_CODESIGN)
```
laforel: HDFLASH_CODESIGN shouldn't appear in HVT — HVT is upstream/OSS and shouldn't reference downstream/internal products by name in its public build contract. I suspect this works today because the parent build (Flash, via HVTEx, via Components) sets HDFLASH_CODESIGN before add_subdirectory(hvt) and HVT inherits it implicitly. That's clever, but it just hides the coupling instead of removing it.

Could we rename to HVT_CODESIGN (HVT-local, consumer-agnostic)? On the Flash side, it's a one-line mirror in the parent CMake before add_subdirectory:
`set(HVT_CODESIGN ${HDFLASH_CODESIGN})`

HVT already has one HdFlash mention (in framePass.cpp, runtime check against the public render-delegate display name). That one is borderline-acceptable because it's runtime, against a publicly-exposed name. A build-time switch named after an internal product is a step further than that precedent, and I don't think we want to grow the pattern.

DDoS: +1

source/CMakeLists.txt
```diff
@@ -116,6 +116,27 @@ endif()
 
 #---------------------------------------------------------------------------------------------------
 
+# macOS codesign
+if (APPLE)
+    if (DEFINED codeSignID)
```
laforel: codeSignID is the only camelCase, unprefixed variable in this block; everything else is HVT_*. 
Simplest end state is to drop the intermediate variable entirely and just have callers set HVT_CODESIGN_IDENTITY directly, defaulting to "-" (ad-hoc). Something like:
```
if (APPLE)
    set(HVT_CODESIGN_IDENTITY "-" CACHE STRING "Codesign identity for HVT dylibs on macOS (\"-\" = ad-hoc)")
    if (HVT_CODESIGN)
        add_custom_command(TARGET ${_TARGET} POST_BUILD
            COMMAND /usr/bin/codesign --force --sign "${HVT_CODESIGN_IDENTITY}" --timestamp=none
                    "$<TARGET_FILE:${_TARGET}>"
            COMMENT "Codesigning ${_TARGET} ($<TARGET_FILE:${_TARGET}>)"
            VERBATIM
        )
    endif()
endif()
```
Not mandatory if you'd rather keep this PR small, the rename in the comment bellow is the main issue IMO.

source/CMakeLists.txt
hodoulp: > HDFLASH_CODESIGN shouldn't appear in HVT — HVT is upstream/OSS and shouldn't reference downstream/internal products by name in its public build contract. I suspect this works today because the parent build (Flash, via HVTEx, via Components) sets HDFLASH_CODESIGN before add_subdirectory(hvt) and HVT inherits it implicitly. That's clever, but it just hides the coupling instead of removing it.
> 
> Could we rename to HVT_CODESIGN (HVT-local, consumer-agnostic)? On the Flash side, it's a one-line mirror in the parent CMake before add_subdirectory: `set(HVT_CODESIGN ${HDFLASH_CODESIGN})`

Fixed ie., no Flash here.

hodoulp: > HVT already has one HdFlash mention (in framePass.cpp, runtime check against the public render-delegate display name). That one is borderline-acceptable because it's runtime, against a publicly-exposed name. A build-time switch named after an internal product is a step further than that precedent, and I don't think we want to grow the pattern.

That's the limitation of all of this as `Storm` specific code paths are deep in the Hydra code.

laforel (approved):

PR #168, opened 2026-05-14
Pr/use pxr ns macro

dj-mcg:
We need to use the macros defined in USD's pxr/pxr.h.in in order for code to build cleanly internally at Pixar.


hodoulp (approved):

hodoulp:
Thanks for your contribution @dj-mcg

PR #173, opened 2026-06-04
Camera related updates

AdamFelt:
This PR broadens camera support.

1) Add linearExposureScale to the camera.  This is part of the HdCameraSchema and is used to control camera exposure.  This can be used in conjunction with Photographic Camera Controls, or used directly.

2) Support directly setting the cameraPath so an application can specify which camera to use for a render pass.  HVT was previously always using the Free camera that is generated internally.

In order to support linearExposureScale the camera storage was switched from the internal HdxFreeCameraSceneDelegate to a Hydra 2.0 workflow where cameras are defined using the HdCameraSchema and added to a SceneIndex.  Care was taken to ensure that if a camera did not change no changes notifications are sent.  The LightingManager also needed to be updated to receive the right transforms from the free camera or the user supplied camera path.

- [X ] **I have signed the Contributor License Agreement (CLA)** ([Corporate](../CLA/ADSKFormCorpContribAgmtforOpenSource.docx) or [Individual](../CLA/ADSKFormIndContribAgmtforOpenSource.docx))
- [ X] My code follows the project's coding standards
- [ X] My changes generate no new warnings or errors


sebastienberube-adsk (approved):
Approved, but consider moving `BuildCameraPrimDataSource()` and `_CameraPrimMatches()` out of framePass.cpp (suggestion framePassUtils.h/.cpp).
Perhaps some of the code or logic added to `FramePass:GetRenderTasks()` could also be moved out of framePass.cpp.

Other than that, I like the option for a user-supplied scene camera and the fact that we no longer use the `HdxFreeCameraSceneDelegate`.

PR #175, opened 2026-06-04
AGP-456 - Fix default paths for wasm

hodoulp:
## Description

The pull request fixes the default path for wasm.

### Tests Performed
- [x] Tested with different render delegates

## Documentation

## Checklist

### DIFF
--- a/source/utils/pathUtils.cpp
+++ b/source/utils/pathUtils.cpp
@@ -58,6 +58,11 @@ namespace
 
 std::string CleanPath(std::string const& path)
 {
+    if (path.empty() || path == "/")
+    {
+        return path;
+    }
+
     const auto clean = std::filesystem::absolute(path);
     return std::filesystem::canonical(clean).string();
 }
@@ -82,6 +87,9 @@ std::string GetExecutable()
         uint32_t size = EXE_PATH_SIZE;
         _NSGetExecutablePath(path, &size);
         exe = path;
+#elif defined(__EMSCRIPTEN__)
+        // No native executable on WASM.
+        exe = "/";
 #elif defined(__linux)
         char path[EXE_PATH_SIZE] = "";
         // Linux read process exe.
@@ -122,14 +130,19 @@ std::filesystem::path GetDefaultResourceDirectory()
 #elif defined(__ANDROID__)
     const char* localAppPath = std::getenv("LOCAL_APP_PATH");
     std::filesystem::path assetsPath = localAppPath ? localAppPath : "";
-    return assetsPath.append(
-        "Resources"); // TODO: OGSMOD-7219
-                      // Standardize usage of lowercase "resource" folder (gizmos/assets).
+    // TODO: OGSMOD-7219 Standardize usage of lowercase "resource" folder (gizmos/assets).
+    return assetsPath.append("Resources");
+#elif defined(__EMSCRIPTEN__)
+    // WASM has no native executable directory. By convention, application
+    // resources are mounted in the Emscripten virtual filesystem at /Resources/
+    // via `--preload-file <src>@/Resources/...` link options. Callers that need
+    // a different layout must use SetResourceDirectory() before any
+    // Get*Path() call.
+    return std::filesystem::path("/Resources");
 #else
     std::filesystem::path exePath = GetCurrentProcessDirectory();
-    return exePath.append(
-        "Resources"); // TODO: OGSMOD-7219
-                      // Standardize usage of lowercase "resource" folder (gizmos/assets).
+    // TODO: OGSMOD-7219 Standardize usage of lowercase "resource" folder (gizmos/assets).
+    return exePath.append("Resources");
 #endif
 }
 
--- a/source/utils/pathUtils.cpp
+++ b/source/utils/pathUtils.cpp
@@ -58,6 +58,9 @@ namespace
 
 std::string CleanPath(std::string const& path)
 {
+    // Bypass canonicalisation for inputs that std::filesystem::canonical
+    // either can't resolve (empty path) or that would round-trip to themselves
+    // anyway (filesystem root).
     if (path.empty() || path == "/")
     {
         return path;
@@ -115,6 +118,9 @@ std::string GetExecutable()
     return std::filesystem::path(exePath).parent_path();
 }
 
+// TODO: OGSMOD-7219 Standardize usage of lowercase "resource" folder (gizmos/assets).
+[[maybe_unused]] constexpr const char* kResourceSubdir = "Resources";
+
 } // anonymous namespace
 
 std::filesystem::path GetDefaultResourceDirectory()
@@ -128,7 +134,7 @@ std::filesystem::path GetDefaultResourceDirectory()
     return std::filesystem::path([resourcePath UTF8String]);
 #endif
 #elif defined(__ANDROID__)
-    const char* localAppPath = std::getenv("LOCAL_APP_PATH");
+    const char* localAppPath         = std::getenv("LOCAL_APP_PATH");
     std::filesystem::path assetsPath = localAppPath ? localAppPath : "";
     // TODO: OGSMOD-7219 Standardize usage of lowercase "resource" folder (gizmos/assets).
     return assetsPath.append("Resources");
@@ -138,11 +144,9 @@ std::filesystem::path GetDefaultResourceDirectory()
     // via `--preload-file <src>@/Resources/...` link options. Callers that need
     // a different layout must use SetResourceDirectory() before any
     // Get*Path() call.
-    return std::filesystem::path("/Resources");
+    return std::filesystem::path("/") / kResourceSubdir;
 #else
-    std::filesystem::path exePath = GetCurrentProcessDirectory();
-    // TODO: OGSMOD-7219 Standardize usage of lowercase "resource" folder (gizmos/assets).
-    return exePath.append("Resources");
+    return GetCurrentProcessDirectory().append(kResourceSubdir);
 #endif
 }
 
@@ -153,7 +157,7 @@ std::filesystem::path GetDefaultMaterialXDirectory()
     // same folder as the executable.
     // For iOS, Frameworks is under the root of bundle, like "Application Bundle/Frameworks"
 #if defined(__APPLE__)
-    NSBundle* bundle = [NSBundle mainBundle];
+    NSBundle* bundle       = [NSBundle mainBundle];
     NSString* resourcePath = [bundle resourcePath];
 #if !TARGET_OS_IPHONE
     resourcePath = [resourcePath stringByDeletingLastPathComponent];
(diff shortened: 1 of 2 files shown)


source/utils/pathUtils.cpp
laferra-adsk, line 168: From AI analysis:

H1 — GetDefaultMaterialXDirectory() on WASM is now "/libraries" (likely still wrong; partial fix)

GetExecutable() now returns "/" on WASM (pathUtils.cpp:90-92). GetCurrentProcessDirectory() returns std::filesystem::path("/").parent_path() which by the C++ filesystem spec is "/". GetDefaultMaterialXDirectory() (pathUtils.cpp:149-166) falls through the #else branch and returns GetCurrentProcessDirectory() = "/". The single caller in source/engine/hgiInstance.cpp:111 does:

`            GetDefaultMaterialXDirectory().append("libraries").string().c_str(), true);`

So on WASM, PXR_MTLX_STDLIB_SEARCH_PATHS is now set to "/libraries". The PR title says "Fix default paths for wasm" (plural) but only the resource path is addressed. If MaterialX libraries are mounted under /Resources/libraries (which is the natural convention given the new resource convention), this remains broken — just broken in a new way.

Proposed fix — add a parallel __EMSCRIPTEN__ branch in GetDefaultMaterialXDirectory():

```
std::filesystem::path GetDefaultMaterialXDirectory()
{
#if defined(__APPLE__)
    NSBundle* bundle = [NSBundle mainBundle];
    NSString* resourcePath = [bundle resourcePath];
#if !TARGET_OS_IPHONE
    resourcePath = [resourcePath stringByDeletingLastPathComponent];
#endif
    NSString* frameworkPath = [resourcePath stringByAppendingPathComponent:@"Frameworks"];
    return std::filesystem::path([frameworkPath UTF8String]);
#elif defined(__EMSCRIPTEN__)
    // WASM: MaterialX libraries are expected next to other resources in the
    // Emscripten virtual filesystem. Keep this aligned with the
    // GetDefaultResourceDirectory() convention.
    return std::filesystem::path("/Resources");
#else
    return GetCurrentProcessDirectory();
#endif
}
```

(Or whatever convention the consuming WASM build actually uses — at a minimum, an explicit branch with a comment, not a silent fallback to "/".)

If the intent is to handle MaterialX in a follow-up, the commit message and PR description should explicitly say so.

test/tests/testPathUtils.cpp
```diff
@@ -71,3 +71,29 @@ TEST(TestPathUtils, GetShaderPath_AppendsCorrectly)
...
+
+TEST(TestPathUtils, EmscriptenDefaultResourceDirectory)
+{
+    // SetResourceDirectory has been called by other tests; 
+    // always reset to default.
+    hvt::SetResourceDirectory({});
+    EXPECT_EQ(hvt::GetResourceDirectory(), std::filesystem::path("/Resources"));
+}
+
+TEST(TestPathUtils, EmscriptenDefaultIsCwdIndependent)
```
laferra-adsk, line 91: From AI analysis:

M1 — EmscriptenDefaultIsCwdIndependent mutates global cwd without restoring it

Issues:

- Process-wide cwd is permanently changed to /tmp after this test runs. Any subsequent test that depends on cwd (or any framework teardown that writes a report relative to cwd) is affected. Test ordering becomes load-bearing.
- std::filesystem::current_path("/tmp") will throw std::filesystem::filesystem_error if /tmp does not exist in the Emscripten MEMFS. There is no guarantee that an HVT consumer's test build mounts /tmp.
- After this test, the next test that calls SetResourceDirectory({}) followed by GetResourceDirectory() continues to get /Resources, so the assertion mechanics are fine — but the global side-effect is a latent footgun.

Proposed fix — save/restore cwd, and create the directory defensively:

```
TEST(TestPathUtils, EmscriptenDefaultIsCwdIndependent)
{
    const auto savedCwd = std::filesystem::current_path();

    hvt::SetResourceDirectory({});
    const auto before = hvt::GetResourceDirectory();

    std::error_code ec;
    std::filesystem::create_directory("/tmp", ec); // best-effort; ignore "already exists"
    std::filesystem::current_path("/tmp");
    hvt::SetResourceDirectory({});
    const auto after = hvt::GetResourceDirectory();

    EXPECT_EQ(before, after);
    EXPECT_EQ(after, std::filesystem::path("/Resources"));

    std::filesystem::current_path(savedCwd);
}
```

source/utils/pathUtils.cpp
```diff
@@ -58,6 +58,11 @@ namespace
 
 std::string CleanPath(std::string const& path)
 {
+    if (path.empty() || path == "/")
```
laferra-adsk, line 64: From AI analysis:

M2 — Behavioral regression in CleanPath("") on Linux readlink failure

The Linux branch of GetExecutable() (pathUtils.cpp:93-103) leaves exe empty if readlink("/proc/self/exe", ...) returns -1. Pre-PR, the subsequent exe = CleanPath(exe) (pathUtils.cpp:107) would resolve the empty string to cwd via absolute("") → canonical(...). Post-PR, the new early return preserves exe as empty.

Downstream:

- GetCurrentProcessDirectory() returns path("").parent_path() = "".
- GetDefaultResourceDirectory() on Linux returns "" / "Resources" = "Resources" (a non-absolute path).
- This would break the GetDefaultResourceDirectory_IsAbsolute test (testPathUtils.cpp:31-35) on Linux when readlink fails — though that failure mode is exceedingly rare in practice.

This is a behavioral change that the commit does not call out. Acceptable in practice (the prior behavior of silently returning cwd was arguably worse), but worth either:

- a one-line comment in CleanPath explaining why empty is returned as-is, or
- handling the readlink-failure branch explicitly (e.g. log + set exe = "/" or similar) so the empty path never propagates.

Proposed comment-only fix:

```
std::string CleanPath(std::string const& path)
{
    // Bypass canonicalisation for inputs that std::filesystem::canonical
    // either can't resolve (empty path) or that would round-trip to themselves
    // anyway (filesystem root). This also avoids the prior, surprising
    // behaviour where CleanPath("") silently resolved to the current working
    // directory.
    if (path.empty() || path == "/")
    {
        return path;
    }

    const auto clean = std::filesystem::absolute(path);
    return std::filesystem::canonical(clean).string();
}
```

test/tests/testPathUtils.cpp
```diff
@@ -71,3 +71,29 @@ TEST(TestPathUtils, GetShaderPath_AppendsCorrectly)
 
     hvt::SetResourceDirectory(saved);
 }
+
+#if defined(__EMSCRIPTEN__)
+
+TEST(TestPathUtils, EmscriptenDefaultResourceDirectory)
+{
+    // SetResourceDirectory has been called by other tests;
```
laferra-adsk: trailing space on this line

test/tests/testPathUtils.cpp
```diff
@@ -71,3 +71,29 @@ TEST(TestPathUtils, GetShaderPath_AppendsCorrectly)
...
 }
+
+#if defined(__EMSCRIPTEN__)
+
+TEST(TestPathUtils, EmscriptenDefaultResourceDirectory)
+{
+    // SetResourceDirectory has been called by other tests; 
+    // always reset to default.
+    hvt::SetResourceDirectory({});
+    EXPECT_EQ(hvt::GetResourceDirectory(), std::filesystem::path("/Resources"));
```
laferra-adsk: From AI analysis:

L3 — EmscriptenDefaultResourceDirectory would benefit from an is_absolute() assertion

The whole point of this fix is to make the default absolute and cwd-independent. The platform-agnostic GetDefaultResourceDirectory_IsAbsolute (testPathUtils.cpp:31-35) already covers absoluteness, so this is just a polish nit — but explicitly asserting is_absolute() here would document intent.

```
TEST(TestPathUtils, EmscriptenDefaultResourceDirectory)
{
    hvt::SetResourceDirectory({});
    const auto& dir = hvt::GetResourceDirectory();
    EXPECT_TRUE(dir.is_absolute());
    EXPECT_EQ(dir, std::filesystem::path("/Resources"));
}
```

source/utils/pathUtils.cpp
```diff
@@ -122,14 +130,19 @@ std::filesystem::path GetDefaultResourceDirectory()
 #elif defined(__ANDROID__)
     const char* localAppPath = std::getenv("LOCAL_APP_PATH");
     std::filesystem::path assetsPath = localAppPath ? localAppPath : "";
-    return assetsPath.append(
```
laferra-adsk, line 125: From AI analysis => this is a code cleanup suggestion

L4 — Pre-existing duplication around .append("Resources")

After the reformat, the Android branch (pathUtils.cpp:130-134) and the default-#else branch (pathUtils.cpp:142-145) both duplicate the same .append("Resources") and the same // TODO: OGSMOD-7219 ... comment.

Optional refactor — pull the "Resources" literal into a constexpr and the TODO comment to a single header above the function so future renames touch one place:

```
// TODO: OGSMOD-7219 Standardize usage of lowercase "resource" folder (gizmos/assets).
static constexpr const char* kResourceSubdir = "Resources";

std::filesystem::path GetDefaultResourceDirectory()
{
#if defined(__APPLE__)
    ...
#elif defined(__ANDROID__)
    const char* localAppPath = std::getenv("LOCAL_APP_PATH");
    std::filesystem::path assetsPath = localAppPath ? localAppPath : "";
    return assetsPath.append(kResourceSubdir);
#elif defined(__EMSCRIPTEN__)
    // ... explanatory comment ...
    return std::filesystem::path("/") / kResourceSubdir;
#else
    return GetCurrentProcessDirectory().append(kResourceSubdir);
#endif
}
```

Skip if you'd rather keep this commit minimal — flag for a follow-up cleanup PR.

laferra-adsk (changes requested):
Suggested changes. One more note, the comment in pathUtils.h for GetDefaultResourceDirectory() is now stale. The doxygen note no longer accurately describes WASM (returns a fixed virtual-filesystem path, not "executable directory appended with..."). Suggest:

```
/// Returns the default resource directory.
/// \note On native platforms the directory is built from the executable
/// directory; on Android, from $LOCAL_APP_PATH; on Emscripten/WASM, it is a
/// fixed virtual-filesystem path (/Resources). Use SetResourceDirectory() to
/// override.
std::filesystem::path GetDefaultResourceDirectory();
```

test/tests/testPathUtils.cpp
hodoulp: Good catch but weak proposal in case of failure. A better fix is to create a `guard`.

test/tests/testPathUtils.cpp
hodoulp: Yes and I improved the test a little bit because it must an absolute path on all plaforms.

source/utils/pathUtils.cpp
hodoulp: Do not change the code because the expected default path for now. We will revisit that later if needed.

laferra-adsk (approved):

PR #181, opened 2026-06-15
KBSH-1269 - Add the Outline Rendering Tasks

julia-b-grenier:
## Description

The pull request adds new public rendering tasks: `outlineMaskTask`, `outlineOverlayTask`, `outlinePrimIdsTask`. These tasks can be use to outline prims with specific customizable styles. It also adds dedicated unit tests validating the behavior and a HowTo test showing how the tasks can be used.

### Changes Made

- Add the tasks `outlineMaskTask`, `outlineOverlayTask`, `outlinePrimIdsTask`
- Create the unit tests to cover all three tasks in `testOutlineTasks.cpp`
- Add the HowTo test in `howTo20_UseOutlineTasks.cpp` to explain the usage of the tasks

## Testing

Added tests for `outlineMaskTask`, `outlineOverlayTask`, `outlinePrimIdsTask` that can be run using
`ctest --preset debug -R TestOutlineTasks` and `ctest --preset debug -R useOutlineTasks`

### Test Configuration
- **Platform(s) tested**: Windows 11
- **Build configuration(s)**: Debug
- **Render delegate(s)**: 
- **OpenUSD version**: 

### Tests Performed

## Documentation

## Checklist


include/hvt/tasks/outline/outlineMaskTask.h
```diff
@@ -0,0 +1,417 @@
...
+
+    PXR_NS::HdRenderIndex* _renderIndex;
+
+    PXR_NS::HgiTextureHandle _outputTexture;
+
+    PXR_NS::HgiBufferHandle _overlayIdValuesBuffer;
+    PXR_NS::HgiBufferHandle _hoverIdValuesBuffer;
+    PXR_NS::HgiBufferHandle _activeIdValuesBuffer;
+
+    PXR_NS::HdStGLSLProgramSharedPtr _computeProgram;
```
julia-b-grenier, line 394: Originally, this variable as well as others were static and define in the .cpp file like so:
```
// Persistent shader cache to prevent using the HdSt shader cache
// which gets garbage collected when ApplyPendingUpdates() called
// from SetTime (during rendering) removes rPrims from the render
// index.
static HdStGLSLProgramSharedPtr s_cachedComputeProgram;
static uint64_t s_cachedComputeProgramHash = 0;
static HgiResourceBindingsSharedPtr s_cachedResourceBindings;
static HgiSamplerHandle s_cachedSampler;
static bool s_samplerInitialized = false;
static uint64_t s_cachedResourceBindingsHash = 0;
static HgiComputePipelineSharedPtr s_cachedPipeline;
static uint64_t s_cachedPipelineHash = 0;
static int s_taskInstanceCount = 0;
```
However, to make this task more consistent with the other tasks in HVT, these static variables have now become members of the OutlineMaskTask class. This change was validated with Claude. Indeed, it is better to have these variables as members instead of static variables. Claude's conclusion:

The member variable design is strictly better than the static design for OutlineMaskTask. The original motivation for the statics — preventing HdSt's shader cache from garbage-collecting the compiled program during ApplyPendingUpdates() — is fully satisfied by member shared_ptr variables, because the task instance outlives any Garbage Collection pass (tasks are not rPrims and are not removed by ApplyPendingUpdates()), so the member shared_ptr keeps the program's ref-count above zero just as effectively as a static. The only scenario where statics would add value over members is a multi-instance case where multiple OutlineMaskTask objects share one compiled shader across viewports — but the real application creates exactly one task instance per viewport and never recreates it per frame, so s_taskInstanceCount only ever moves 0 → 1 → 0 and the shared cache is never actually shared. On top of providing no real benefit, the static design introduced three concrete thread-safety hazards: the plain int s_taskInstanceCount is not atomic, the check-then-cleanup in the destructor is a race if two instances are destroyed concurrently, and the cache-miss/compile/store pattern has a TOCTOU race that can leak a compiled GPU program. Member variables carry none of these risks, are simpler to reason about, tie resource lifetimes cleanly to object lifetime via RAII, and produce identical runtime behavior in the one-instance-per-viewport application that actually exists.

laferra-adsk (approved):
LGTM! Thanks

test/howTos/howTo20_UseOutlineTasks.cpp
```diff
@@ -0,0 +1,317 @@
...
+    ((outlinePrimIdsTaskBase, "outlinePrimIdsTaskBase"))
+    ((outlinePrimIdsTaskOverlay, "outlinePrimIdsTaskOverlay"))
+    ((outlinePrimIdsTaskDefault, "outlinePrimIdsTaskDefault"))
+    ((outlineMaskTask, "outlineMaskTask"))
+    ((outlineOverlayTask, "outlineOverlayTask"))
+);
+
+} // namespace
+
+#if defined(__APPLE__)
```
hodoulp, line 87: Why?

test/howTos/howTo20_UseOutlineTasks.cpp
```diff
@@ -0,0 +1,317 @@
...
+HVT_TEST(howTo, DISABLED_useOutlineTasks)
+#else
+HVT_TEST(howTo, useOutlineTasks)
+#endif
+{
+    if (GetParam() == HgiTokens->Vulkan)
+    {
+        // Vulkan backend render arbitrary fails.
+        GTEST_SKIP() << "Skipping test for the Vulkan backend.";
```
hodoulp, line 96: Why?

test/howTos/howTo20_UseOutlineTasks.cpp
```diff
@@ -0,0 +1,317 @@
...
+    //
+    // Each OutlinePrimIdsTask exports two task-context entries keyed by its prefix:
+    //   "outline<Prefix>PrimIdsTexture"
+    //   "outline<Prefix>DepthTexture"
+    // OutlineMaskTask reads all six entries (two per prefix) and runs a GPU compute
+    // shader that produces a single RGBA mask texture.
+
+    auto& taskManager = sceneFramePass->GetTaskManager();
+
+    GfVec2i currentBufSize { 0, 0 };
```
hodoulp: Why? The sceneFramePass already holds the buffer size!

test/howTos/howTo20_UseOutlineTasks.cpp
```diff
@@ -0,0 +1,317 @@
...
+        init.enabled      = true;
+        init.bufferPrefix = "Base";
+        init.collection   = HdRprimCollection(HdTokens->geometry,
+            HdReprSelector(HdReprTokens->smoothHull), SdfPath("/Root/Selected"));
+
+        taskManager->AddTask<hvt::Outline::OutlinePrimIdsTask>(
+            _tokens->outlinePrimIdsTaskBase, init, makePrimIdsCommit);
+    }
+
+    // "Overlay" — manipulator/on-top objects (disabled here; no overlay prims in this scene).
```
hodoulp, line 202: Do we have to know upfront? 
if yes, that seems way too much because they can be dynamic (based on user settings for example).

source/tasks/outline/outlinePrimIdsTask.cpp
```diff
@@ -0,0 +1,721 @@
...
+            TF_CODING_ERROR("Cannot initialize render pass state: picking shader path is empty");
+            return rps;
+        }
+        auto pickGlslfx = std::make_shared<HioGlslfx>(shaderPath, HioGlslfxTokens->defVal);
+        extendedState->SetRenderPassShader(std::make_shared<HdStRenderPassShader>(pickGlslfx));
+    }
+    return rps;
+}
+
+TfToken _GetOutlineTextureToken(std::string const& prefix, char const* suffix)
```
hodoulp: [Minor] usually `prefix` is before

source/tasks/outline/outlineOverlayTask.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+PXR_NAMESPACE_USING_DIRECTIVE
+
+namespace HVT_NS::Outline
+{
+
+TF_DEFINE_PRIVATE_TOKENS(_tokens,
+    ((outlineMaskTexture, "outlineMaskTexture"))((shaderNoBlur, "OutlineOverlayTask::NoBlur"))(
+        (shaderBlur3x3, "OutlineOverlayTask::Blur3x3"))(
+        (shaderBlur5x5, "OutlineOverlayTask::Blur5x5")));
```
hodoulp: [Minor] usually one entry per line to make it more redeable

test/howTos/howTo20_UseOutlineTasks.cpp
hodoulp: I found. The challenge was already noticed with `display_primId_AOV` i.e., disabled on macOS / Metal.

source/tasks/outline/outlineMaskTask.cpp
```diff
@@ -0,0 +1,1359 @@
...
+    HVT_OUTLINE_MASK_PARAMS,
+    HVT_OUTLINE_MASK_RESOURCES,
+    HVT_OUTLINE_MASK_SHADERCODE
+);
+
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
+#pragma GCC diagnostic ignored "-Wc++20-extensions"
+#endif
```
llaforest: These are Clang-only warnings; on real GCC this is the `-Werror=pragmas` error. Switch the guard to `__clang__` and use `#pragma clang diagnostic`:

```suggestion
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
#pragma clang diagnostic ignored "-Wc++20-extensions"
#endif
```

source/tasks/outline/outlineMaskTask.cpp
```diff
@@ -0,0 +1,1359 @@
...
+    TF_DEBUG_ENVIRONMENT_SYMBOL(HVT_OUTLINE_MASK_CACHE, "outline mask resource caching");
+    TF_DEBUG_ENVIRONMENT_SYMBOL(HVT_OUTLINE_MASK_PARAMS, "outline mask configuration params");
+    TF_DEBUG_ENVIRONMENT_SYMBOL(
+        HVT_OUTLINE_MASK_SHADERCODE, "outline mask shader code before and after compilation");
+    TF_DEBUG_ENVIRONMENT_SYMBOL(HVT_OUTLINE_MASK_RESOURCES, "outline mask resources");
+}
+
+#if defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
```
llaforest: Matching closing block:

```suggestion
#if defined(__clang__)
#pragma clang diagnostic pop
#endif
```

llaforest (commented):
The Linux GPU CI build (`ci-test.yaml`) is failing to compile these two new outline task files:

```
source/tasks/outline/outlineMaskTask.cpp:43:32: error: unknown option after '#pragma GCC diagnostic' kind [-Werror=pragmas]
source/tasks/outline/outlinePrimIdsTask.cpp:39:32: error: unknown option after '#pragma GCC diagnostic' kind [-Werror=pragmas]
ninja: build stopped: subcommand failed.
```

**Root cause:** `-Wgnu-zero-variadic-macro-arguments` and `-Wc++20-extensions` are **Clang-only** warning names, but they're guarded with `#if defined(__GNUC__)`. Clang also defines `__GNUC__`, so this compiles fine locally (macOS/Clang), but the Linux runner uses real **GCC**, which doesn't recognize those options and — with the repo's `-Werror=pragmas` — turns it into a hard error.

**Fix:** guard with `__clang__` and use `#pragma clang diagnostic` instead (GCC never emits those warnings, so it needs no suppression). Same applies to the matching `pop` blocks. One-click suggestions inline below. 👇

source/tasks/outline/outlinePrimIdsTask.cpp
```diff
@@ -0,0 +1,721 @@
...
+    HVT_OUTLINE_PRIM_IDS_PARAMS,
+    HVT_OUTLINE_PRIM_IDS_RESOURCES,
+    HVT_OUTLINE_PRIM_IDS_VALIDATE
+);
+
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
+#pragma GCC diagnostic ignored "-Wc++20-extensions"
+#endif
```
llaforest: Same issue here — Clang-only warnings guarded by `__GNUC__`:

```suggestion
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
#pragma clang diagnostic ignored "-Wc++20-extensions"
#endif
```

source/tasks/outline/outlinePrimIdsTask.cpp
```diff
@@ -0,0 +1,721 @@
...
+{
+    TF_DEBUG_ENVIRONMENT_SYMBOL(
+        HVT_OUTLINE_PRIM_IDS_PARAMS, "outline primIds configuration params");
+    TF_DEBUG_ENVIRONMENT_SYMBOL(HVT_OUTLINE_PRIM_IDS_RESOURCES, "outline primIds resources");
+    TF_DEBUG_ENVIRONMENT_SYMBOL(HVT_OUTLINE_PRIM_IDS_VALIDATE, "outline primIds validate results");
+}
+
+#if defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
```
llaforest: Matching closing block:

```suggestion
#if defined(__clang__)
#pragma clang diagnostic pop
#endif
```

julia-b-grenier:
Thank you @llaforest for the suggestions and for the explanation about the `__clang__ `and `__GNUC__ `distinction.

test/howTos/howTo20_UseOutlineTasks.cpp
julia-b-grenier: Yes, issues around `display_primId_AOV` with MacOS and Metal were documented, so I decide to be safe and also disable it for my test as well. Is that okay?

test/howTos/howTo20_UseOutlineTasks.cpp
julia-b-grenier: This test was failing for Vulkan and I saw that this is how they are disabled in other similar code. I believe that maybe this test should also be disabled so it does not arbitrary fails.

test/howTos/howTo20_UseOutlineTasks.cpp
julia-b-grenier: I believe this is indeed a mistake, thank you for noticing. I will fix this.

test/howTos/howTo20_UseOutlineTasks.cpp
julia-b-grenier: No we do not have to know upfront, it can definitely be dynamic. For instance, when an application that consumes those tasks update the outline task parameters, it can verify if there are overlay paths(its the application mandate to keep track of the default/base/overlay paths and feed them to the outline tasks) and then dynamically decide if there are overlays or not:
```
bool hasOverlay = !overlayPaths.empty();
outlineOverlayPrimIdsTaskParams.enabled = hasOverlay;
``` 
That comment only describes the static setup of this particular how-to scene, which happens to have no overlay prims; it's not a constraint of the tasks.

source/tasks/outline/outlineOverlayTask.cpp
julia-b-grenier: Good catch, will fix this thank you Patrick!

source/tasks/outline/outlinePrimIdsTask.cpp
julia-b-grenier: Good point, I will rename it to category as it reflects one of the category Default/Base/Overlay.

julia-b-grenier:
Thank you @hodoulp for your suggestions. 

Here is a summary of the commits I made following your suggestions:
- This commit addresses the findings you reported to me that Cursor gave you: [KBSH-1269 Fix outline task defects and remove dead API](https://github.com/Autodesk/hydra-viewport-toolbox/pull/181/commits/b63b0a4f221feed4a390441ae78a6750b141858e)
- The following commits are about the suggestions made in this code review: [Rename prefix for category for the function _GetOutlineTextureToken](https://github.com/Autodesk/hydra-viewport-toolbox/pull/181/commits/2a29ec3cc4f218289e3151a566b5acb6ee0e45b7), [Use the sceneFramePass buffer size instead](https://github.com/Autodesk/hydra-viewport-toolbox/pull/181/commits/5688ee99a6961161959879649d55737f4cc12a30), 
[Improve readability of TF_DEFINE_PRIVATE_TOKENS](https://github.com/Autodesk/hydra-viewport-toolbox/pull/181/commits/88cb26e54b4a423f7ddc5b0d6fc9d949d16ec2b5)

Please let me know what you think of my answers to the remaining questions that you asked in this code review. 

Thank you so much!

laforel:
Make sure you squash and merge, do not rebase those 13 commits!

hodoulp (approved):
Great Job

hodoulp:
⚠️ Cannot be merged before having `HVTEx` and `Components` green.

PR #183, opened 2026-07-09
AGP-517 Support wider range of OpenUSD versions (24.11 -> 26.05) with dual SI/SD task backends

sebastienberube-adsk:
## Description

To support a wider range of OpenUSD versions, this PR proposes a common interfaces to support both SceneIndex (SI) and SceneDelegate (SD) task data and prim handling.

image

image

### Test Configuration
- **Platform(s) tested**: Mac, Windows, Linux
- **Build configuration(s)**: Debug, Release
- **Render delegate(s)**: Storm (OpenGL, Metal, Vulkan)
- **OpenUSD version**: 24.11 (Windows only), 26.05 (Mac, Windows, Linux)

### Tests Performed

- [x] Tested with different render delegates

## Documentation

## Checklist


source/engine/taskBackendFactory.cpp
```diff
@@ -0,0 +1,132 @@
...
+#endif
+    auto* sd = dynamic_cast<TaskSDBackend*>(taskBackend.get());
+    TF_VERIFY(sd, "TaskBackend is neither SI nor SD");
+    if (sd)
+    {
+        return std::make_unique<LightingPrimSDBackend>(
+            pRenderIndex, sd->GetSyncDelegate());
+    }
+    return nullptr;
+}
```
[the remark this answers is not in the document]
sebastienberube-adsk: This was resolved, TF_FATAL_ERROR is called when task backend is invalid.

source/engine/taskBackendFactory.cpp
```diff
@@ -0,0 +1,132 @@
...
+#endif
+    auto* sd = dynamic_cast<TaskSDBackend*>(taskBackend.get());
+    TF_VERIFY(sd, "TaskBackend is neither SI nor SD");
+    if (sd)
+    {
+        return std::make_unique<RenderBufferPrimSDBackend>(
+            pRenderIndex, sd->GetSyncDelegate());
+    }
+    return nullptr;
+}
```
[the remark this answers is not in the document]
sebastienberube-adsk: This was resolved, TF_FATAL_ERROR is called when task backend is invalid.

source/engine/sd/taskSDBackend.cpp
```diff
@@ -0,0 +1,135 @@
...
+    spec.sdCreate(_renderIndex, _syncDelegate.get(), taskId);
+
+    // Store the initial task parameters in the scene delegate.
+    _syncDelegate->SetValue(taskId, HdTokens->params, spec.params);
+}
+
+void TaskSDBackend::RemoveTask(SdfPath const& taskId)
+{
+    _renderIndex->RemoveTask(taskId);
+}
```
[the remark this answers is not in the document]
sebastienberube-adsk: This is a non-issue, the FramePass is responsible for cleaning up everything. See FramePass::Uninitialize(), this is the same as the previous logic.

source/engine/sd/renderBufferPrimSDBackend.cpp
```diff
@@ -0,0 +1,131 @@
...
+void RenderBufferPrimSDBackend::InsertRenderBuffer(
+    SdfPath const& id, HdRenderBufferDescriptor const& desc, size_t msaaSampleCount)
+{
+    _pRenderIndex->InsertBprim(HdPrimTypeTokens->renderBuffer, _syncDelegate.get(), id);
+    _syncDelegate->SetValue(id, _tokens->renderBufferDescriptor, VtValue(desc));
+    _syncDelegate->SetValue(
+        id, HdStRenderBufferTokens->stormMsaaSampleCount,
+        VtValue(desc.multiSampled ? msaaSampleCount : static_cast<size_t>(1)));
+    _pRenderIndex->GetChangeTracker().MarkBprimDirty(id, HdRenderBuffer::DirtyDescription);
+}
```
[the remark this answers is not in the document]
sebastienberube-adsk: Although this could be changed for the sake of consistency between SI and SD, we make the choice to keep the SD implementation as close as possible as it was in the previous implementation before the TaskManager was migrated to the SceneIndex (commit [7bfc0f1]), where size_t was used.

See:
https://github.com/Autodesk/hydra-viewport-toolbox/blob/a347db92880004418ac092f869ac88e601a63b5a/source/engine/renderBufferManager.cpp#L670-L675

source/engine/sd/renderBufferPrimSDBackend.cpp
```diff
@@ -0,0 +1,131 @@
...
+                bprimDirty = true;
+            }
+        }
+
+        if (msaaSampleCountChanged)
+        {
+            _syncDelegate->SetValue(
+                id, HdStRenderBufferTokens->stormMsaaSampleCount, VtValue(msaaSampleCount));
+            bprimDirty = true;
+        }
```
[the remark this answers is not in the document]
sebastienberube-adsk: Although this could be changed for the sake of consistency between SI and SD, we make the choice to keep the SD implementation as close as possible as it was in the previous implementation before the TaskManager was migrated to the SceneIndex (commit [7bfc0f1]), where size_t was used.

See:
https://github.com/Autodesk/hydra-viewport-toolbox/blob/a347db92880004418ac092f869ac88e601a63b5a/source/engine/renderBufferManager.cpp#L887-L892

source/engine/sd/syncDelegate.h
```diff
@@ -0,0 +1,151 @@
...
+    /// \id The task unique identifier. e.g. "/framepass_main_0/colorCorrectionTask"
+    /// \key The parameter key. e.g. "params", "renderTags", "collection"
+    /// \return An empty VtValue if the value does not exist, or the value if it does.
+    PXR_NS::VtValue GetValue(PXR_NS::SdfPath const& id, PXR_NS::TfToken const& key) const;
+    
+    /// Gets a value pointer from the sync delegate with the specified ID and key, if it exists.
+    /// \id The task unique identifier. e.g. "/framepass_main_0/colorCorrectionTask"
+    /// \key The parameter key. e.g. "params", "renderTags", "collection"
+    /// \return A nullptr if the value does not exist, or the value if it does.
+    const PXR_NS::VtValue* GetValuePtr(PXR_NS::SdfPath const& id, PXR_NS::TfToken const& key) const;
```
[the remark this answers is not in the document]
sebastienberube-adsk: This is out of scope for the current PR, this file is from a previous revision (https://github.com/Autodesk/hydra-viewport-toolbox/blob/a347db92880004418ac092f869ac88e601a63b5a/include/hvt/engine/syncDelegate.h#L101-L105) and was just brought back to life to support the SD backend.

source/engine/taskBackendFactory.h
```diff
@@ -0,0 +1,65 @@
...
+// functions below can be declared. Their definitions are private (source/engine/) and not
+// installed, so external callers can only pass the resulting handles back into HVT (e.g. through
+// FramePass), not dereference or destroy them directly.
+class FramePassCamera;
+class LightingPrimBackend;
+class RenderBufferPrimBackend;
+
+using FramePassCameraPtr         = std::unique_ptr<FramePassCamera>;
+using LightingPrimBackendPtr     = std::unique_ptr<LightingPrimBackend>;
+using RenderBufferPrimBackendPtr = std::unique_ptr<RenderBufferPrimBackend>;
```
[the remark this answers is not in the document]
sebastienberube-adsk: Non-issue. These classes are not meant to be be fully public and destroyed by an external user.
The same strategy is used for the following classes, this is just following the same pattern (LightingManager is not public):
https://github.com/Autodesk/hydra-viewport-toolbox/blob/a347db92880004418ac092f869ac88e601a63b5a/include/hvt/engine/framePass.h#L56-L58

test/tests/testEngineHelpers.cpp
```diff
@@ -340,9 +343,10 @@ struct TaskManagerTestFixture
 
     TaskManagerTestFixture(SdfPath const& uid, HdRenderIndex* pRenderIndex)
     {
-        retainedSceneIndex = HdRetainedSceneIndex::New();
-        pRenderIndex->InsertSceneIndex(retainedSceneIndex, SdfPath::AbsoluteRootPath());
-        taskManager = std::make_unique<hvt::TaskManager>(uid, pRenderIndex, retainedSceneIndex);
+        std::shared_ptr<hvt::TaskBackend> taskBackend = hvt::CreateTaskBackend(
+            pRenderIndex, SdfPath::AbsoluteRootPath(), /*useLegacySceneDelegate=*/false);
+        taskManager        = std::make_unique<hvt::TaskManager>(uid, pRenderIndex, taskBackend);
```
hodoulp: Can you simplify the code?
```
        std::shared_ptr<hvt::TaskBackend> taskBackend = hvt::CreateTaskBackend(
            pRenderIndex, SdfPath::AbsoluteRootPath(), /*useLegacySceneDelegate=*/false);
        taskManager        = std::make_unique<hvt::TaskManager>(uid, pRenderIndex, taskBackend);
```
be replaced by
```
        taskManager        = std::make_unique<hvt::TaskManager>(uid, pRenderIndex, /*useLegacySceneDelegate=*/false);
```
or
```
        taskManager        = std::make_unique<hvt::TaskManager>(uid, pRenderIndex);
```

include/hvt/engine/taskBackend.h
```diff
@@ -0,0 +1,136 @@
...
+    /// Type-erased task creator for the scene-delegate (SD) backend.
+    using SdTaskCreatorFn = std::function<void(PXR_NS::HdRenderIndex* renderIndex,
+        PXR_NS::HdSceneDelegate* sceneDelegate, PXR_NS::SdfPath const& taskId)>;
+
+    /// SD backend: inserts the task into the render index via the scene delegate.
+    SdTaskCreatorFn sdCreate;
+
+#if HVT_ENABLE_SI_TASK_BACKEND
+    /// SI backend: legacy task factory used to instantiate the HdTask from the scene index.
+    PXR_NS::HdLegacyTaskFactorySharedPtr siFactory;
```
hodoulp, line 94: As this is in a public header, does a user of the library could/need to use this struct parameter? if yes that's unfortunate as he will need ifdef in his code. Can that be transparent?

source/engine/taskBackendFactory.h
```diff
@@ -0,0 +1,65 @@
...
+
+#include <hvt/api.h>
+#include <hvt/engine/taskBackend.h>
+
+#include <pxr/imaging/hd/renderIndex.h>
+#include <pxr/usd/sdf/path.h>
+
+#include <memory>
+
+namespace HVT_NS
```
hodoulp, line 24: Why all these backends must be public? At first glance that seems to be some implementation 'details'. Can that be private?

hodoulp (commented):
That's a first review.
- Add an md file in docs to explain the rational, design, etc.

hodoulp:
Run `Cursor` or `Claude`. In my context `Cursor` highlights several points to improve. I suggest to have a look to the `high` ones.

```
AGPHVT-122 — Dual SI/SD TaskBackend Support
Branch: berubes/feature/AGPHVT-122-SI-SD-Support

Summary
Adds dual Hydra backend support so HVT works on USD 24.11 (scene-delegate / SD) and newer USD versions (scene-index / SI, default). Introduces a TaskBackend abstraction shared by TaskManager, RenderBufferManager, and LightingManager, with SI/SD implementations and a runtime/build-time backend selector.

Key changes
- New TaskBackend interface plus SI (TaskSIBackend) and SD (TaskSDBackend) implementations
- New UseSceneIndex() / SetUseSceneIndex() and HVT_USE_SCENE_INDEX env var
- FramePass captures backend at init and routes tasks, lights, buffers, and camera through the selected backend
- SI path: single retained scene index (same model as before)
- SD path: restored SyncDelegate-based scene delegate path for USD 24.11
- Factory helpers: CreateTaskBackend(), CreateFramePassCamera(), etc.
- Tests updated for new TaskManager API; SD parity image tests added on USD 25.05+
- USD version guards in visualizeAovTask / wboitResolveTask for older HGI APIs

Review findings (recommended follow-ups)

High
1. TaskSDBackend::SetValue returns true for invalid task paths / unsupported keys (SI backend validates correctly; SD-only on 24.11 is affected)
2. SD SyncDelegate entries not cleared on task/light/render-buffer removal (stale data risk on path reuse)
3. SD free camera ignores linearExposureScale (SI supports it via HdCameraSchema)

Medium 
4. CreateFramePassCamera silently falls back to SD if backend type mismatches (lighting/buffer factories fatal-error instead)
5. TaskSDBackend::Uninitialize() is a no-op (asymmetric vs SI)
6. SetUseSceneIndex() is process-global and not thread-safe
7. Confusing naming: _useSceneIndex vs useLegacySceneDelegate
8. Breaking public API: TaskManager and RenderBufferManager constructors changed

Low 
9. docs/framepass.md still documents old architecture
10. Duplicate FramePassCameraPtr typedef
11. SD validation tests disabled on USD < 25.05
12. No explicit test for linearExposureScale SD vs SI difference

Strengths
- Clean abstraction and per-FramePass backend consistency
- Good macOS/clang visibility fix for HdLegacyTaskFactory
- ScopedSceneIndexMode for test isolation
- Correct USD version compatibility shims
```

include/hvt/engine/taskBackend.h
sebastienberube-adsk: That is indeed unfortunate, but this needs to be visible from any user code that creates custom tasks.

This is because `PXR_NS::HdMakeLegacyTaskFactory<T>` is templated, so the definition of the factory needs to be known by the caller of `TaskManager::AddTask<T>` (which is also public and templated). We can't fully hide the `HdLegacyTaskFactory<T>` from the user, since the actual factory type can't be known in advance.

For the record, when the TaskManager was migrated to a Scene Index (https://github.com/Autodesk/hydra-viewport-toolbox/pull/153) , this factory was also made publics but _without ifdefs_, breaking compilation with older versions of USD.

**Why we should keep it this way**
There would be ways around this, but that would require modifying the TaskManager interface, which we want to avoid:
e.g.
`taskManager->AddTask<HdxRenderTask>(name, params, commitFn);`
to
`taskManager->AddTask(name, MakeTaskCreateInfo<HdxRenderTask>(VtValue(params)), commitFn);`

The modification above would allow the template to be removed from `TaskManager::AddTask`, but would still require to expose the templated function `MakeTaskCreateInfo<HdxRenderTask>`, which is worse in my opinion. The taskBackend itself could then be made private, but we would need to extract `MakeTaskCreateInfo<T>` and make a subset of the backend public to support task creation by external code. I don't see a clean way around this limitation. At least the current solution preserves the existing TaskManager interface.

test/tests/testEngineHelpers.cpp
sebastienberube-adsk: In short: yes it is possible to remove the task backend from the **tests**.

I have added a way for the `TaskManager` to create the `TaskBackend` automatically when no arguments for it are provided.

I am however keeping the `taskBackend` argument in the `FramePass`, since that backend is shared will all other SI/SD managers.

source/engine/taskBackendFactory.h
sebastienberube-adsk: The backend factory, specifically, can be made private. This header was kept public to allow more thorough testing in hvt\test\tests\testTaskManager.cpp.

However, I just made changes to testTaskManager.cpp to avoid explicit task backend creation without loss of coverage. (Using the terminal SceneIndex from the RenderIndex instead of the HdRetainedSceneIndex from the task backend - this is possible because the terminal scene index dispatches the GetPrim method to the actual si implementation). As a bonus, these tests can now support both SI/SD, which wasn't the case before.

sebastienberube-adsk:
@hodoulp 
> Review findings (recommended follow-ups)
> 
> High
> 1. TaskSDBackend::SetValue returns true for invalid task paths / unsupported keys (SI backend validates correctly; SD-only on 24.11 is affected)
> 2. SD SyncDelegate entries not cleared on task/light/render-buffer removal (stale data risk on path reuse)
> 3. SD free camera ignores linearExposureScale (SI supports it via HdCameraSchema)
> 
> Medium 
> 4. CreateFramePassCamera silently falls back to SD if backend type mismatches (lighting/buffer factories fatal-error instead)
> 5. TaskSDBackend::Uninitialize() is a no-op (asymmetric vs SI)
> 6. SetUseSceneIndex() is process-global and not thread-safe
> 7. Confusing naming: _useSceneIndex vs useLegacySceneDelegate
> 8. Breaking public API: TaskManager and RenderBufferManager constructors changed
> 
> Low 
> 9. docs/framepass.md still documents old architecture
> 10. Duplicate FramePassCameraPtr typedef
> 11. SD validation tests disabled on USD < 25.05
> 12. No explicit test for linearExposureScale SD vs SI difference

1. Unlike the SI backend, the SD backend SetValue can't fail. It will always set the value for the given path. That is fine.
2. The SD SyncDelegate, along with all its data, is deleted in FramePass::Uninitialize(). If a task or render buffer is removed while the FramePass is still alive, it is true the data remains in the SD SyncDelegate, but that is not an issue in the current usage context with the FramePass. This data is lightweight, and there are mechanisms in the RenderBufferManager and LightingManager to refresh it. That is exactly how it was working before the migration of the TaskManager to SI. I consider this point to be a non-issue and out of scope for the current set of changes.
3. `linearExposureScale` is indeed not supported by SD camera, that is expected.

4. Fixed, an error message was added.
5. SI and SD are somewhat asymmetric by nature, this is fine. 
6. This is fine, SetUseSceneIndex() is expected to be process-global. The only use case for this is tests.
7. Fixed. _useSceneIndex was renamed to _useLegacySceneDelegate.
8. Partly flixed: A constructor without TaskBackend argument was added for the TaskManager. For the RenderBufferManager however, it is currently not expected to be created by entities other than the FramePass, so this a non-issue.

9. Fixed, doc updated.
10. FramePassCameraPtr typedef duplication allows for lighter includes (forward declaration). That is a voluntary choice, and I don't want to create new header files just to centralize this typedef.
11. SD does not have 100% test coverage, that is a choice that was made to keep test times under control.
12. Non-issue (linearExposureScale unsupported with SD).

docs/framepass.md
```diff
@@ -6,50 +6,74 @@ A `FramePass` is the top-level unit of rendering in the HVT engine. It owns a re
 
 ## Architecture
 
-Each `FramePass` creates a shared `HdRetainedSceneIndex` and delegates prim management to three specialised managers:
+HVT supports two rendering backends, selected at `FramePass` construction time:
+
+| Backend | Abbreviation | USD Requirement | How prims are managed |
+|---|---|---|---|
+| **Scene Index** | SI | USD >= 25.05 (`HdLegacyTaskSchema`) | Prims live in an `HdRetainedSceneIndex` inserted into the render index |
+| **Scene Delegate** | SD | Any USD version | Prims are inserted/removed through an `HdSceneDelegate` (sync delegate) |
```
hodoulp, line 14: You can perhaps add a comment mentioning that the `scene index` is the recommended case for newer USD versions (as `scene delegate` is slowly deprecated).

include/hvt/engine/taskBackend.h
```diff
@@ -47,6 +48,33 @@ PXR_NAMESPACE_CLOSE_SCOPE
...
+/// HVT supports two rendering backends:
+///  - scene-index (SI): the Hydra 2.0 retained-scene-index based path (default);
+///  - scene-delegate (SD): the legacy HdSceneDelegate based path.
+///
+/// The backend is selected at FramePass construction time. The default is SI (this function
+/// returns false), overridable through the \c HVT_USE_LEGACY_SCENE_DELEGATE environment variable
+/// and at runtime via SetUseLegacySceneDelegate().
+///
+/// \note On USD versions that lack HdLegacyTaskSchema (e.g. 24.11, i.e.
+/// HVT_ENABLE_SI_TASK_BACKEND is true) the SI backend cannot be built, so this always returns
```
hodoulp: There are two env. variables `HVT_USE_LEGACY_SCENE_DELEGATE` & `HVT_ENABLE_SI_TASK_BACKEND` ?

test/tests/testTaskManager.cpp
```diff
@@ -201,7 +201,7 @@ HVT_TEST(TestTaskManager, taskmgr_integration)
 
 // Retained scene index inspection requires the legacy task schema (USD >= 25.05); the fixture
 // always uses the SD backend below that, so this test is disabled in that case.
-#if HVT_ENABLE_SI_TASK_BACKEND
+#if !HVT_ENABLE_SI_TASK_BACKEND
```
hodoulp: I suggest to have same name for env. variable, C++ define, etc. to avoid confusion.

hodoulp (approved):

docs/framepass.md
sebastienberube-adsk: I believe the **legacy** terminology speaks for itself. The **Scene Index backend** is also the default option when creating a **FramePass**, if you don't specifically call `UseLegacySceneDelegate()`.

include/hvt/engine/taskBackend.h
sebastienberube-adsk: No.
Only `HVT_USE_LEGACY_SCENE_DELEGATE` is an environment variable.
`HVT_ENABLE_SI_TASK_BACKEND ` is just a macro based on OpenUSD version, telling if the Scene Index backend is supported.

I agree this can be confusing, so I just renamed `HVT_ENABLE_SI_TASK_BACKEND` to `HVT_SI_TASK_BACKEND_SUPPORTED`.

In short, even if `HVT_SI_TASK_BACKEND_SUPPORTED` symbol is true, you can still force the use of the SD backend by setting `HVT_USE_LEGACY_SCENE_DELEGATE` env var to true.

test/tests/testTaskManager.cpp
sebastienberube-adsk: I just renamed `HVT_ENABLE_SI_TASK_BACKEND` to `HVT_SI_TASK_BACKEND_SUPPORTED`, for clarity.
 
 The `HVT_USE_LEGACY_SCENE_DELEGATE` env var is a choice.
 The `HVT_SI_TASK_BACKEND_SUPPORTED` is a symbol used as a precompiler directive to exclude headers, structs or classes that don't exist in older versions of OpenUSD where SI-based tasks were not yet supported.

PR #184, opened 2026-07-15
[NOM-5453] Allocate fresh depth buffer when pass's sample count differs from the chained input

LukPopp0:
## Description

RenderBufferManager::SetRenderOutputs reuses the previous pass's buffer when the renderer is unchanged, in order to preserve sub-pixel information. That reuse assumed the chained buffer always has the same multisample state as the pass consuming it.

This is not true when a pass's sample count differs from the pass before it. A multisampled overlay pass chained after a resolved, single-sampled pass would inherit the single-sampled buffer. A single-sampled buffer attachment cannot be bound alongside a multisampled color target, so the overlay silently loses multisampling (or fails validation, depending on backend).

This PR makes buffer reuse conditional on a matching sample count. On a same-renderer sample-count mismatch, a fresh buffer is allocated at the pass's own sample count instead of reusing the mismatched one. The matching-sample-count and different-renderer paths are left unchanged.

### Changes Made

- Fold a multisample-state check into the `inputFound` predicate, so it applies uniformly to every AOV (color and depth) rather than being special-cased for depth.
- Only reuse the chained buffer when input.buffer->IsMultiSampled() matches the pass's multiSampled descriptor.
- On a same-renderer mismatch, force allocation of a fresh buffer at the pass's sample count and skip the copy (the resolved single-sampled buffer has no meaningful per-sample copy into a multisampled target, and overlay passes clear their buffers before drawing).
- Leave the different-renderer path untouched, so existing configurations behave exactly as before..
 
## Testing

Verified with a multisampled overlay pass (4x MSAA) chained after a resolved, single-sampled pass. Before the change the overlay inherited single-sampled buffer and could not enable MSAA; after the change it receives a correctly multisampled buffer and renders antialiased. Confirmed the single-sampled (non-MSAA) path and the existing cross-renderer copy path are unaffected.

### Test Configuration
- **Platform(s) tested**: macOS (Metal)
- **Build configuration(s)**: Release
- **OpenUSD version**: 26.05

### Tests Performed

## Documentation

## Checklist

### DIFF
--- a/source/engine/renderBufferManager.cpp
+++ b/source/engine/renderBufferManager.cpp
@@ -668,7 +668,14 @@ bool RenderBufferManager::Impl::SetRenderOutputs(TfToken const& outputToVisualiz
                     depthDesc  = desc;
                     depthInput = input;
 
-                    if (inputFound)
+                    // Reuse the previous pass's depth buffer unless its multisample state
+                    // mismatches this pass. An MSAA overlay pass chained after a resolved
+                    // single-sampled pass would otherwise inherit single-sampled depth, which
+                    // cannot be attached alongside a multisampled color target.
+                    const bool depthMultisampleMismatch =
+                        input.buffer && (input.buffer->IsMultiSampled() != desc.multiSampled);
+
+                    if (inputFound && !depthMultisampleMismatch)
                     {
                         // If the renderer remains the same, we don't want to copy the depth buffer.
                         // The existing depth buffer will continue to be used.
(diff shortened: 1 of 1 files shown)


source/engine/renderBufferManager.cpp
sebastienberube-adsk: I am curious if there is a particular reason why MSAA match check is reserved only for depth buffers.

Would a more general solution like this also work?
```
const bool depthMultisampleMismatch = input.buffer && (input.buffer->IsMultiSampled() != desc.multiSampled);
inputFound = (rendererName == input.rendererName ) && !depthMultisampleMismatch;
```

sebastienberube-adsk:
I left a minor comment for you to consider, but the core idea makes sense.
I'll come back and approve once the comment is answered, thanks!

source/engine/renderBufferManager.cpp
LukPopp0: @sebastienberube-adsk I like this change, applied it and tested it with my setup. I also updated the PR description accordingly to be more general instead of focused on the depth buffer.

sebastienberube-adsk (approved):

hodoulp:
Sorry @LukPopp0 to jump late and thanks @sebastienberube-adsk for reviewing the work.
However, I have some questions:
- Did someone validate the change against `HVTEx` and `Components` ? 
- Can you write a unit test demonstrating that it works as expected?

source/engine/renderBufferManager.cpp
```diff
@@ -662,7 +662,15 @@ bool RenderBufferManager::Impl::SetRenderOutputs(TfToken const& outputToVisualiz
...
-                inputFound = (rendererName == input.rendererName);
+                // Reuse the previous pass's buffer only when it has the same multisample state as
+                // this pass. An MSAA pass chained after a resolved single-sampled pass would
+                // otherwise inherit a single-sampled buffer, which cannot be attached alongside a
+                // multisampled target. On a mismatch treat the input as not found so a fresh
+                // buffer is allocated.
+                const bool multisampleMismatch =
+                    input.buffer && (input.buffer->IsMultiSampled() != desc.multiSampled);
+                inputFound = (rendererName == input.rendererName) && !multisampleMismatch;
+
```
hodoulp, line 673: Good catch

sebastienberube-adsk:
@hodoulp 
> Did someone validate the change against HVTEx and Components ?

I have ingested that change into my [HVT PR-183](https://github.com/Autodesk/hydra-viewport-toolbox/pull/183/changes) which was tested with the full suite.

PR #186, opened 2026-07-22
KBSH-1269 - Add the OutlineManager

julia-b-grenier:
## Description

The pull request adds a new public feature-level manager: `outlineManager`. `OutlineManager` is a wrapper around the outline rendering tasks added in Phase 1 (`outlineMaskTask`, `outlineOverlayTask`, `outlinePrimIdsTask`). It owns the task IDs, AOV texture bindings and the ordering of the tasks inside the frame pass, so consumers no longer have to rediscover the correct wiring themselves. State is push-based: hosts call `SetInputs()` / `SetStyle()` whenever their selection or style changes, and the manager caches so a call with unchanged inputs is a no-op. It also adds dedicated unit tests and a HowTo test showing how the manager can be used.

`OutlineManager` deliberately knows nothing about how the host tracks selection (signal/observer/scene-index/etc.); the host wires its own selection source to `SetInputs()`. Two new public contracts express that boundary: `OutlineInputs` (the selected / lead / hover / overlay / exclude path buckets plus `isHoverSelected`) and `OutlineStyle` (colors, softness, blur and the default-outline toggle).

### Changes Made

- Add `hvt::Outline::OutlineManager`, a feature-level wrapper around the outline tasks (`Install`, `SetInputs`, `SetStyle`, `GetCacheStats`)
- Add the public `OutlineInputs` and `OutlineStyle` contracts consumed by the manager
- Create the unit tests covering the manager in `testOutlineManager.cpp`
- Add the HowTo test in `howTo21_UseOutlineManager.cpp` to explain the usage of the manager
- Update `outline.md` documentation to describe the manager and the input/style contracts
- Rename "active" to "lead" across the public params, the compute shader, the storage buffers and the debug output (see Breaking changes)
- Fix a GPU resource leak in OutlineMaskTask: resource-bindings and compute-pipeline
  handles are non-owning, so they are now released explicitly (in cleanup and on cache
  miss) instead of relying on shared_ptr wrappers that never freed them
- Outline overlay now preserves destination alpha. OutlineOverlayTask composites with separate alpha blending — color stays (SrcAlpha, OneMinusSrcAlpha), alpha changes from (One, Zero) to (One, OneMinusSrcAlpha), matching the existing convention in wboitResolveTask. The previous state overwrote the color AOV's alpha with the mask's alpha (~0 outside outline edges), zeroing scene alpha across the full-screen quad; that alpha is forwarded to the burnthrough/ViewCube passes and the final canvas composite. This is a rendering behavior change and accounts for the four updated image baselines (`howTo/useOutlineTasks`, `outline_renderEnabled`, `outline_renderBlurModes`, `outline_renderNonBasePrefix`): a channel-wise diff of each against its pre-PR version confirms RGB is bit-identical and only the alpha channel moved, on every pixel. Those four are the only pre-existing baselines this PR modifies (`git diff --diff-filter=M 64d5619 HEAD -- 'test/data/baselines/*'`); the other ten baselines in the diff are new files, so they have no prior version to compare against. RenderingUtils::compareImages loads both PNGs as RGBA and diffs all four channels, so the baselines cover the new behavior going forward.
- Keep the outline tasks' dirty bits set when the params fetch fails. Clearing them was permanent — TaskManager::SetTaskValue emits no dirty locator for an unchanged value, so no further DirtyParams arrived and the task was stranded on stale params. Leaving them set lets a later re-sync retry, matching the adjacent initialization-failure path; a latched TF_WARN reports the failure once per streak instead of once per frame. The non-Storm and disabled early-outs still clear, and now record that DirtyParams is the only bit these tasks consume.
- Report OutlinePrimIdsTask initialization failures instead of inferring them, so the task can no longer claim success when it has not initialized. A failed initialization disables the task so Prepare() / Execute() cannot dereference a null render-pass state, and leaves the dirty bits set so a later sync retries. Three changes make the retry sound: it tests every resource rather than the render pass alone (a pass created before the state failed used to make the next attempt skip initialization entirely and report success, leaving the state null); _CreateAovBindings() returns bool and clears a partially-built binding set, which previously survived for the life of the task because a non-empty _aovBuffers kept the caller from re-entering — this is the one failure reachable in practice, an AOV allocation failing under GPU memory pressure; and each step is guarded separately, so a retry re-attempts only what is missing. The render-pass and render-pass-state diagnostics are latched, since that retry re-runs on every sync, and the latch is released as each resource comes up so a pass that recovers cannot silence a state failure that follows.
- Resolve leadPath / hoverPaths / overlayPaths to primitive IDs in OutlineMaskTask::Prepare() instead of _Sync(). HdTask permits Hydra to skip Sync for a task with no dirty bits, and the manager emits byte-identical params on quiet frames, so the resolution depended on unspecified behavior; Prepare() is guaranteed for every task on every execute. A HdChangeTracker::GetRprimIndexVersion() gate means it re-resolves only when the result can have changed, rather than on every frame. Hosts need not notify the task of scene changes.
- For shader paths, keep generic_u8string() and add a begin()/end() copy into std::string

### Breaking changes

Both are public-API breaks in `hvt::Outline`; there is no deprecation shim.

- `active` -> `lead` rename: `OutlineMaskTaskParams::activePath` -> `leadPath`, `activeIdValues` -> `leadIdValues`, `OutlineMaskStyleParams::activeIdsCount` -> `leadIdsCount`, and the matching GLSL constant / storage-buffer names in `outlineMask.glslfx`.
- `OutlineMaskTask::SetVisualizationMode` removed. It wrote the mode on the task's local params copy and then marked DirtyParams, so the next _Sync refetched the stored params and overwrote it — the write never took effect, and a task has no route to the TaskManager that owns the stored params. Select the mode through `OutlineStyle::maskVisualizationMode` + `OutlineManager::SetStyle()`, or through `OutlineMaskTaskParams` when driving the raw tasks.

## Testing

Added tests for `outlineManager` that can be run using
`ctest --preset debug -R TestOutlineManager` and `ctest --preset debug -R useOutlineManager`

`OutlineManager` is verified to be a faithful refactor of the manual five-task wiring, not merely a
comparable one: `test/data/baselines/howTo/useOutlineManager.png` (howTo21, the wrapper) and
`test/data/baselines/howTo/useOutlineTasks.png` (howTo20, the raw tasks) are the **same git blob**
(`5bbe99a6`), so the two paths render pixel-identically.

```
git ls-tree HEAD test/data/baselines/howTo/useOutlineManager.png \
                 test/data/baselines/howTo/useOutlineTasks.png
```

Note on coverage: the outline render tests are `DISABLED_` on `__APPLE__` (primId rendering is
non-deterministic there) and `GTEST_SKIP`ed on Vulkan, so the image-baseline coverage is desktop GL.

### Test Configuration
- **Platform(s) tested**: Windows 11, MacOS Tahoe 26.5.1
- **Build configuration(s)**: Debug
- **Render delegate(s)**: Storm (OpenGL). The outline tasks are Storm-only by design — `_Enabled()` returns `_isStormRenderer` — and the image baselines are desktop GL, per the coverage note above.
- **OpenUSD version**: 26.5 (`vcpkg.json`)

### Tests Performed

## Documentation

## Checklist


hodoulp:
`Claude` found:

```
Real problems

1. Base collection can contain duplicate root paths — outlineManager.cpp:377-379
SdfPathVector roots = s.inputs.selectedPaths;
roots.insert(roots.end(), s.inputs.hoverPaths.begin(), s.inputs.hoverPaths.end());
No dedup. The overlap case isn't hypothetical — it's the documented isHoverSelected state ("a single hovered candidate is already in the selection set", outlineManager.h:121), so the hovered path is in both buckets. HdRprimCollection::SetRootPaths sorts but doesn't unique, and HdPrimGather::Filter gives no guarantee for repeated include paths. Best case: that prim is rasterized twice into the base primId AOV every frame. Fix: sort+unique before SetRootPaths.

2. TF_WARN on a fully supported configuration — outlineManager.cpp:251-257
enableDefaultOutlines is a first-class style knob whose own doc comment says "Disable for performance on heavy scenes" (outlineManager.h:52-54). Turning it off warns. Worse, the warn block sits after params.enabled is computed, so Install() + enableDefaultOutlines=false + empty selection warns even though the entire outline group is inert. The warnedDefaultAlias latch is itself an admission that the message is noise. Either drop it or make it TF_DEBUG. Two of the new tests (outline_maskTextureFallbackWhenDefaultDisabled, outline_perBucketEnabledFlags) trip it.

3. Input change-detection is detached from the struct it compares — outlineManager.cpp:408-413 / outlineManager.h:107
OutlineStyle has operator==; OutlineInputs doesn't, so SetInputs() hand-compares six fields in a different file. Add a field to OutlineInputs and this silently keeps returning "unchanged": no generation bump, the per-task CollectionCache never invalidates, and outlines render against stale paths — with no compile error. Give OutlineInputs an operator== next to its fields and call it.

4. Dangling framePass + permanent install latch — outlineManager.cpp:112, 152-156, 169
Impl::framePass is a raw pointer set once and never cleared, and Install() refuses a second call forever. The header's lifetime design is entirely about the manager dying first (weak_ptr no-op); the reverse ordering — frame pass recreated on device loss or swapchain rebuild — leaves framePass dangling with no way to re-install. docs/outline.md asserts "The frame pass must outlive the manager" but nothing in the header states or enforces it. Either document it as a hard precondition in the header, or add a detach so the manager can be re-installed.

5. The five task names and six texture names are duplicated string literals — outlineManager.cpp:44-65
The manager hardcodes "outlineBasePrimIdsTask" etc. instead of calling OutlinePrimIdsTask::GetToken("Base") / OutlineMaskTask::GetToken() / OutlineOverlayTask::GetToken(), which produce the same strings. Same for _PrimIdsTextureName/_DepthTextureName, which re-implement _GetOutlineTextureToken (anonymous namespace, outlinePrimIdsTask.cpp:102). testOutlineManager.cpp asserts both spellings agree, which is exactly the tell that they can diverge. Use the public GetToken() accessors and promote the texture-name helper to a shared header.

6. The dirty-bit fix was applied to one task out of three — outlineOverlayTask.cpp:74
Correct fix. But the identical early-return-without-clearing paths remain in outlinePrimIdsTask.cpp:296-299 and outlineMaskTask.cpp:546-549, 554-557, 567-570. In particular OutlineMaskTask::_Sync's if (!_Enabled()) return; means the mask task re-syncs every frame forever under any non-Storm renderer — the exact condition the overlay comment describes.

7. enableDefaultOutlines defaults to true — outlineManager.h:54
A bare Install() with no selection immediately costs a whole-scene primId render pass, a compute dispatch, and a fullscreen composite every frame, plus a visible faint outline on everything. That's a surprising default for a wrapper marketed as the easy entry point.

Minor

- outlineManager.cpp:134 — mutable CacheStats stats is unnecessary; _impl is a non-const shared_ptr, sn mutate through it regardless.
- CacheStats::maxCollectionSize/avgCollectionSize measure input path counts, not collection sizes (a path expands to a subtree). The name will mislead whoever reads the debug output.
- outlineMaskTask.cpp:503-536 — _CleanupAovBindings() now destroys the bindings/pipeline but leaves _rineHash stale. Harmless today (the handle null-check guards first) but fragile; zero them.
- generic_u8string() → generic_string() (3 sites) is the right C++20 fix, but it's lossy for non-ASCII install paths on Windows.
- docs/outline.md:288 — mask.leadPath            =  lost its column alignment in the rename.
```

test/howTos/howTo21_UseOutlineManager.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    // Step 2: Create and install the OutlineManager.
+    //
+    // Install() wires the five internal tasks into the frame pass in the correct
+    // execution order (prim-IDs -> mask -> overlay), sets up AOV texture bindings,
+    // and attaches commit functions that read viewport parameters (render-buffer
+    // size, camera, framing) from the frame pass on every Render() call.
+    //
+    // This single call replaces the five AddTask<>() calls and three commit
+    // lambdas in howTo20_UseOutlineTasks.cpp.
```
hodoulp, line 147: The comment seems outdated

test/data/baselines/outline_renderStyleChange_none.png
hodoulp, line 1: The images are so small that I do not see the differences between `outline_renderStyleChange_none.png‎`,  `outline_renderStyleChange_blur5x5.png` and `outline_renderStyleChange_blur3x3.png`. Can you find a way to increase the images differences?

source/tasks/outline/outlineOverlayTask.cpp
```diff
@@ -145,8 +148,16 @@ void OutlineOverlayTask::Execute(HdTaskContext* ctx)
...
+    // src-over. Color uses (SrcAlpha, OneMinusSrcAlpha) because the source color is not
+    // premultiplied. The alpha channel uses (One, OneMinusSrcAlpha): the source factor is
+    // One (not SrcAlpha) because that is canonical src-over for the alpha channel itself --
+    // using SrcAlpha would square the source alpha and leave antialiased outline edges
+    // slightly under-opaque. The previous (One, Zero) overwrote the destination alpha with
+    // the mask's alpha (~0 except near outline edges), wiping the AOV's alpha across the
+    // full-screen quad.
     _fullscreenShader->SetBlendState(true, HgiBlendFactorSrcAlpha, HgiBlendFactorOneMinusSrcAlpha,
-        HgiBlendOpAdd, HgiBlendFactorOne, HgiBlendFactorZero, HgiBlendOpAdd);
+        HgiBlendOpAdd, HgiBlendFactorOne, HgiBlendFactorOneMinusSrcAlpha, HgiBlendOpAdd);
```
hodoulp, line 172: Good catch

include/hvt/tasks/outline/outlineManager.h
```diff
@@ -0,0 +1,215 @@
...
+        size_t avgCollectionSize = 0;
+    };
+    CacheStats GetCacheStats() const;
+
+private:
+    class Impl;
+    // shared_ptr so the task commit callbacks can hold a weak_ptr and lock()-check it,
+    // making them safe to run after this manager is destroyed (they no-op) -- matching the
+    // weak-pointer pattern used by the built-in HVT tasks.
+    std::shared_ptr<Impl> _impl;
```
hodoulp: Comments:

- If I understand correctly that not a private implementation but rather a private structure holding some data. Name is ambiguous.
- The implementation is clearly an implementation detail that user should never know i.e., that's why there is indirection. 
   - Note: `std::shared_ptr` perhaps better `std::unique_ptr` and `fnComit` can consume a raw pointer. Simplier approach.

source/tasks/outline/outlineManager.cpp
```diff
@@ -0,0 +1,455 @@
...
+// Per-task cache of the derived HdRprimCollection. The collection depends only on the
+// path inputs, so it is rebuilt lazily and reused across commits whose inputs are unchanged
+// (keyed against Impl::inputsGeneration, which SetInputs bumps on every real change).
+struct CollectionCache
+{
+    uint64_t generation = std::numeric_limits<uint64_t>::max();
+    PXR_NS::HdRprimCollection collection;
+};
+
+void _ApplyViewportParams(
```
hodoulp: Name is mabiguous `_ApplyViewportParams()` is `_GetViewportParams()`

source/tasks/outline/outlineManager.cpp
```diff
@@ -0,0 +1,455 @@
...
+    // textures to base (enableDefaultOutlines=false), so the warning fires at most once
+    // per manager instead of once per process. Only touched on the render (commit) thread.
+    bool warnedDefaultAlias = false;
+
+    // Running sum of the per-query collection size across every SetInputs() call (hits and
+    // misses alike). Kept exact so CacheStats::avgCollectionSize is a single integer division
+    // at the end rather than a compounding running mean.
+    size_t collectionSizeSum = 0;
+
+    mutable CacheStats stats;
```
hodoulp: What is that parameter?

test/howTos/howTo21_UseOutlineManager.cpp
julia-b-grenier: I do not think it is outdated. Instead of calling the AddTask and the commit lambdas like it is done in `howTo20_UseOutlineTasks.cpp`, it simply just calls `outline.Install(*sceneFramePass);`. This test is doing the same thing as `howTo20_UseOutlineTasks.cpp`, but it highlights how using the OutlineManager actually makes the usage of the feature much more simpler.

test/data/baselines/outline_renderStyleChange_none.png
julia-b-grenier: I added a camera zoom so it is easier to see the differences.

include/hvt/tasks/outline/outlineManager.h
julia-b-grenier: I renamed the private structure for something less ambiguous and I left a comment explaining the usage of `std::shared_ptr` instead of `std::unique_ptr`.

source/tasks/outline/outlineManager.cpp
julia-b-grenier: Done, thank you for the recommendation.

source/tasks/outline/outlineManager.cpp
julia-b-grenier: I left a comment and removed the mutable.

julia-b-grenier: It was in the original ContourManager and was useful for debugging purposes. Is it okay to keep it in the OutlineManager as well?

julia-b-grenier:
@hodoulp Thank you for the suggestions, I addressed them in this commit(including the suggestions from Claude): [Add suggestions from code review](https://github.com/Autodesk/hydra-viewport-toolbox/pull/186/commits/865ef1d00fc89004a100abcb6792c4038b4cd3ad)

include/hvt/tasks/outline/outlineManager.h
```diff
@@ -194,20 +211,29 @@ class HVT_API OutlineManager
...
+    // pass, so they outlive this manager; the weak_ptr lets a commit that runs after the manager
+    // is destroyed safely no-op instead of dereferencing freed state.
+    //
+    // The built-in HVT managers (RenderBufferManager, LightingManager, SelectionHelper) get the
+    // same callback safety with a plain unique_ptr pimpl because the frame pass owns *them* as
+    // shared_ptr SettingsProviders -- the callbacks weak-reference the manager itself. This
+    // OutlineManager is deliberately host-owned (by value) and decoupled from the frame pass
+    // ownership graph, so there is no manager-level shared_ptr to weak-reference; the shared/weak
+    // boundary is pushed down to this state instead. A raw pointer here could not be liveness-checked.
+    class SharedState;
```
hodoulp, line 250: The explanation is verbose and seems out dated or out of scope.
This variable purpose is quite simple: The class privately holds its state.

Note: The fnCommit methods can use `_state.get()` i.e.., the raw pointer. But this is internal implementation which should hidden from the header.

hodoulp (approved):

include/hvt/tasks/outline/outlineManager.h
julia-b-grenier: You're right, it is indeed way too verbose. I changed it to a concise comment. Thank you Patrick!

julia-b-grenier:
Alain provided several suggestions generated by Claude. These suggestions have been addressed, and the following document summarizes the issue identified in each finding and the approach taken to resolve it.

# KBSH-1269 phase 2 — response to the HVT PR #186 reviews

> **Written:** 2026-08-03
> **PR:** https://github.com/Autodesk/hydra-viewport-toolbox/pull/186 — "KBSH-1269 - Add the OutlineManager"
> **Range covered:** `eb9c7e72` (inclusive) → HEAD `142f1fd`, in `src/Core/Components/HVTEx/externals/source/hvt`
> **Reviews answered:** `KBSH-1269-OutlineManager-HVT.md` (findings **1–14**) and `KBSH-1269-OutlineManager-HVTEx.md` (**H1–H2, M1–M5, N1–N9**)

One row per distinct finding; both reviewers' IDs share a column where they overlap.

## Commits in range

12 outline commits (oldest → newest) plus the `origin/main` merge. The `~1..HEAD` range also pulls in
five unrelated upstream commits that arrived via that merge — `e69315d` (#183), `a3bee48` (#188),
`acd7cde` (#189), `8c1564a` (#187), `07e280b` (#190) — ignore those when reading the range.

| # | SHA | Subject |
|---|---|---|
| 1 | `eb9c7e7` | Correct outline leadPath contract and prune redundant collection roots |
| 2 | `fe4adb3` | Document the alpha coverage of the outline render baselines |
| 3 | `7132624` | Keep outline task dirty bits set when the params fetch fails |
| 4 | `ffb263c` | Split OutlineMaskTask resource cleanup by lifetime |
| 5 | `b4e7091` | Narrow the outline collection builders to OutlineInputs |
| 6 | `4294a90` | Document why the base prim-IDs pass stays enabled with an empty collection |
| 7 | `3b44c31` | Clean up outline task formatting nits |
| 8 | `5dfcce9` | Remove OutlineMaskTask::SetVisualizationMode |
| 9 | `0b1ff20` | Warn when an outline task's params fetch fails |
| 10 | `7d4d152` | Tidy up the outline task API surface and comments |
| 11 | `817c96f` | Intern the outline task-context tokens once |
| 12 | `1738576` | Resolve outline paths to primIds in Prepare(), not _Sync() |
| — | `142f1fd` | Merge `origin/main` |

Five new tests landed across the range: `outline_baseCollectionPrunesDuplicateHoverRoot`,
`...PrunesNestedHoverRoot`, `...KeepsStringPrefixSiblingRoot` (`eb9c7e7`), `outline_renderResize`
(`ffb263c`), `outline_renderVisualizationModes` (`5dfcce9`, replacing a deleted test), and
`outline_renderLeadPicksUpInsertedPrim` (`1738576`).

---

## Fixed

| IDs | Problem | Approach | Where (HEAD) | Commit |
|---|---|---|---|---|
| **1** | `leadPath` contract: the header said "MUST be one of `selectedPaths`", but the doc's own example passes a *descendant*, and the failure mode was a silent no-draw | Reworded to the real rule — leadPath is never rasterized; it recolors only those of its resolved IDs already in the base texture, so member / descendant / spanning-ancestor all work, and only a disjoint leadPath has no effect. Added a per-path de-duped warning when it resolves to zero prim IDs. leadPath still not added to the base roots | `outlineManager.h:104-108`; warn `outlineMaskTask.cpp:711-719` + `_lastWarnedLeadPath` (`outlineMaskTask.h:438-441`); base-roots rationale `outlineManager.cpp:370-375` | `eb9c7e7` |
| **1** *(the "adjacent, lower priority" note)* | Overlapping roots when a child of a selected parent is hovered — the old sort+unique dedup only collapsed exact duplicates | Replaced the dedup with `SdfPath::RemoveDescendentPaths`, which also drops nested roots. **The review's premise was wrong here and the commit says so:** overlap never caused double rasterization (`HdPrimGather` partitions the rprim list into disjoint ranges and gathers each prim once). The real win is normalization — `HdRenderPass::SetRprimCollection` early-outs on collection equality, so hover changes inside an already-selected subtree stop dirtying the collection and re-gathering draw items | `outlineManager.cpp:52-67`; base builder `:368-378`; `docs/outline.md:172`; tests `testOutlineManager.cpp:753`, `:771`, `:787` (incl. the string-prefix-sibling case that must *not* be pruned) | `eb9c7e7` |
| **3 / H2** | `*dirtyBits = Clean` on the `!_GetTaskParams()` path latched a one-off failure permanently — `SetTaskValue` emits no dirty locator for an unchanged value, so no further `DirtyParams` arrives | Dropped the clear so the bits stay set and a later re-sync retries, matching the adjacent `_InitIfNeeded` failure path. The non-Storm and disabled paths still clear, and now record that `DirtyParams` is the only bit these tasks consume. Then added a latched `TF_WARN`, cleared on the next successful fetch, so a persistent failure is not silent | `outlineMaskTask.cpp:586-602`, `outlineOverlayTask.cpp:71-84`, `outlinePrimIdsTask.cpp:305-313`; `_paramsFetchWarned` in all three headers | `7132624` + `0b1ff20` |
| **4** | The non-Storm / disabled dirty-bit clear assumed a renderer switch recreates the tasks — an unverified assumption | Wrote the invariant down at both sites (mask and prim-IDs): a later params update re-dirties the task, and an `HdRenderIndex`'s render delegate is fixed at construction, so switching renderer destroys the task rather than leaving `_isStormRenderer` stale. `1738576` sharpened it further — clearing the bits reports them as *consumed*; it does not stop Hydra syncing the task | `outlineMaskTask.cpp:576-584`, `:615-620`; `outlinePrimIdsTask.cpp` same shape | `ffb263c` (+ `7132624`, refined in `1738576`) |
| **5 / M1** | `_CleanupAovBindings()` released the compute pipeline together with the size-dependent resources, and `_CreateAovBindings()` calls it on **every** render-buffer size change — a drag-resize destroyed and recreated the pipeline once per frame | Split by lifetime: `_CleanupAovResources()` keeps only what depends on buffer size / ID counts; the pipeline moves to `_DestroyPipeline()` and the sampler to `_DestroySampler()`; the destructor calls all three. **Went past the finding**: the sampler is now destroyed through the cached `HdxTask` Hgi instead of `_renderIndex->GetResourceRegistry()->GetHgi()`, which silently leaked it when `_renderIndex` was null; `_samplerInitialized` dropped in favour of the handle's own validity; `_DestroyPipeline()` reused at the `Execute` cache-miss site so the handle is released in one place | `outlineMaskTask.h:334-347`; defs `outlineMaskTask.cpp:507`, `:542`, `:555`; dtor `:170-172`; cache-miss `:1005`. New test `outline_renderResize` (`testOutlineTasks.cpp:693-695`) renders full → half → full against the `outline_renderEnabled` baseline | `ffb263c` + `7d4d152` |
| **6** | `_ValidatePrimIdBuffer` logged "Selected prims in RenderIndex" while printing *all* rprims | Made the string accurate rather than reverting to the old "Active": now **"All prims in RenderIndex"**, since it logs `GetRprimIds()` unfiltered — neither the selected nor the lead prims | `outlinePrimIdsTask.cpp:588-590` | `ffb263c` |
| **7** | `CollectionCache` keyed on `inputsGeneration`, but the builders took `SharedState const&` and so *could* read `style` — no stale read today, but the invariant rested on a comment | Took the structural option over the one-line one: narrowed `collectionFn` to `std::function<HdRprimCollection(OutlineInputs const&)>` so the cache key provably covers everything a builder is handed. Avoids the needless per-style-change collection rebuild that bumping the counter in `SetStyle` would have cost | `outlineManager.cpp:305-308`; lambdas `:368`, `:382`, `:387`; call site `:336`; `SetStyle` unchanged at `:434-441` | `b4e7091` |
| **8 / N7** | The base prim-IDs pass runs with an empty collection in overlay-only / default-only states — correct and tested, but the reason appeared nowhere in the source | Recorded the reason at the call site, tracing it through the mask's Base-texture sampling and its missing-input early-out (disabling the base pass would stop outlines rendering altogether), and naming the two tests that pin it | `outlineManager.cpp:350-358` | `4294a90` |
| **11 / N1** | `docs/outline.md` said the manager owns "the fixed `Overlay > Base > Default` ordering", which reads as task *execution* order | Corrected: the manager owns the task **execution order**; `Overlay > Base > Default` is the layer priority the mask shader resolves, and the phrase now appears only where that is what is meant | `docs/outline.md:19`; priority uses `:54`, `:71` | `3b44c31` |
| **12 / N5** | Trailing whitespace, misaligned `BufferBinding_LeadIdValues` comment, unrelated `TF_DEBUG_ENVIRONMENT_SYMBOL` reformat | Trailing whitespace stripped across the outline sources and headers (zero matches remain), comment column restored. The multi-line `TF_DEBUG_ENVIRONMENT_SYMBOL` form was **kept** and made consistent with the other tasks rather than reverted. Also touched `howTo20_UseOutlineTasks.cpp` and `testOutlineTasks.cpp` | `outlineMaskTask.cpp:99`; `outlinePrimIdsTask.cpp:47-58`, `outlineMaskTask.cpp:52-71` | `3b44c31` |
| **14** | `SetVisualizationMode` wrote the task's *local* params then marked `DirtyParams`, so the next `_Sync` refetched the stored params and overwrote it — the write never took effect, and a task has no route to the `TaskManager` that owns the stored params | Deleted the setter (header, cpp, doc). **Replaced rather than dropped the test**: `outline_maskTaskSetVisualizationMode` only checked the setter ran without error; `outline_renderVisualizationModes` renders a selected box and a nearer sphere once per mode against its own baseline, covering the leg that only shows in pixels — the params value selecting a compute program. It also brackets the camera near/far around the scene instead of using the harness's diameter/100 → diameter×10 range, which compressed all depths to ~0.996 and rendered `VISUALIZE_DEPTH` black | `outline_styleVisualizationModePropagatesAllModes` (`testOutlineManager.cpp:1237`) + `outline_renderVisualizationModes` (`:1611`) + 4 new baselines | `5dfcce9` |
| **M3** | Task-context tokens rebuilt per call from `std::string`, held in non-immortal function-local statics — a concat plus a token intern 4× per frame per task, three tasks installed | Added an immortal `OutlineMaskTextureToken()` accessor next to the name helpers; cached the primId / depth tokens on the task, refreshed only when `bufferPrefix` changes | `outlineTextureNames.h:55-63`; `outlinePrimIdsTask.cpp:469-479`, `:564`; members `outlinePrimIdsTask.h:157`, `:188-191`; consumers `outlineMaskTask.cpp:1230`, `outlineOverlayTask.cpp:126-127` | `817c96f` |
| **M4** | `OutlineManager`'s immovability was a side effect of deleting the copy ops, not a stated decision | Deleted the move ops explicitly, with the rationale in a comment (bound 1:1 to the frame pass it installed into; hold by value or in a `unique_ptr`) | `outlineManager.h:162-171` | `7d4d152` |
| **M5 / 9** | `CacheStats` semantics undocumented: `excludePaths` omitted from the counters though `operator==` compares it; truncating average; `collectionSizeSum` named for a path count | Kept the API and documented what it counts — the highlight buckets only, `excludePaths` deliberately excluded because it filters the default bucket rather than contributing outlined prims, and `avgInputPathCount` truncates. `collectionSizeSum` is not public: it is a private `SharedState` member | `outlineManager.h:219-227`; impl `outlineManager.cpp:408-419`; `collectionSizeSum` `:121-123`; doc mirror `docs/outline.md:157` | `7d4d152` |
| **N3** | `outlineTextureNames.h` missing from `_PRIVATE_HEADER_FILES` | Listed, as `wboitTokens.h` already was | `source/tasks/CMakeLists.txt:56-59` | `7d4d152` |
| **N8** | The overlay blend comment didn't state the destination-alpha assumption behind the colour factors | Recorded that the colour factors assume a non-premultiplied source over an effectively opaque destination | `outlineOverlayTask.cpp:159-172` | `7d4d152` |

## Fixed, but the premise was corrected first

| IDs | Problem as filed | What actually happened | Where | Commit |
|---|---|---|---|---|
| **H1** | Resolved prim IDs go stale: `_Sync` clears its dirty bits and the manager emits byte-identical params, so a render-index change never re-resolves. Proposed `OutlineManager::Invalidate()` + an `OutlineMaskTaskParams::revision` field | **Premise corrected, fix retained.** Instrumenting `_Sync`/`Execute` showed `HdRenderIndex::SyncAll` on the pinned USD syncs clean tasks every frame (12 syncs / 12 frames, nearly all `Clean`), so the IDs could **not** go stale today — but the old code was correct only for as long as an implementation keeps doing what `hd/task.h:72-74` explicitly disclaims, and it re-resolved on 9 of 9 enabled frames. Chose the library-owns-what-it-can-observe route instead of exporting invalidation: resolution moved to `Prepare()` (the phase guaranteed for every task on every execute, and the one `HdTask` designates for querying other prims), gated on `GetRprimIndexVersion()` plus a flag set when `_Sync` replaces the params. **No public API added — no `Invalidate()`, no `revision`, no host wiring, no Falcon or maya-hydra change.** Rationale, measurements and the maya-hydra cross-check (BBox display mode is the case option 1 would have missed): `code-review-phase2.md` | `_ResolvePathsToPrimIds` (`outlineMaskTask.cpp:627`); `_primIdsResolveNeeded` / `_rprimIndexVersion` (`outlineMaskTask.h:429-435`), set at `:612`; new test `outline_renderLeadPicksUpInsertedPrim` (`testOutlineManager.cpp:1779`, `DISABLED_` variant `:1777`) — it pins a reference render against a baseline then compares an insertion-only render against it, so it catches both a gate that stops noticing insertions and a total loss of lead colouring; it cannot exercise the skipped-`Sync` case; `docs/outline.md:211`, `:219`, `:369` | `1738576` |
| **2** | The overlay alpha blend change `(One, Zero)` → `(One, OneMinusSrcAlpha)` is the PR's most consequential edit, moves 4 existing baselines, has no isolating test, and isn't surfaced in the PR description | Blend math confirmed correct and left as is. The coverage objection was answered rather than met with a new unit test: the render tests compare through `RenderingUtils::compareImages`, which loads both PNGs as RGBA and diffs **all four channels**, so the baselines do cover the composite's destination alpha — now noted at both render-test sections. Comment extended (see N8). Unchanged, and still true: the render tests are `DISABLED_` on Apple and `GTEST_SKIP`ed on Vulkan. **The PR-description half of this finding is handled manually** — see below | `testOutlineTasks.cpp:555-556`, `testOutlineManager.cpp:1377-1378`; `outlineOverlayTask.cpp:159-172` | `fe4adb3` |

## No change needed (agreed with the reviewer)

| IDs | Outcome |
|---|---|
| **10 / M2** | `active` → `lead` is a breaking public change, but M2 was withdrawn: the only Phase-1 client is being refactored alongside, so there is no consumer to break. No deprecation shim; belongs in the release notes. |
| **13** | `generic_u8string()` + the `begin()/end()` copy confirmed correct (lossless for non-ASCII install paths, compiles under C++17 and C++20). No code change — but the PR description claims the opposite; see below. |
| **N4** | Withdrawn by the reviewer: `<hvt/api.h>` is required because it pulls in `HVT_NS`. |
| **N9** | `SetStyle` recording no cache stats is intentional; the `CacheStats` doc now states the read-back is inputs-only. |

## Not Fixed or Fixed Manually

| IDs | Item | Comment |
|---|---|---|
| **N2 / 2 / 10** | **The PR description is stale and in two places says the opposite of what shipped.** It still claims "Use `generic_string()` instead of `generic_u8string()`" and "clear dirty bits when params fetch fails" — both inverted by the commits above. It also omits the alpha-blend behavior change that moves 4 baselines, the `active` → `lead` public break, the `SetVisualizationMode` removal, and the `Prepare()` resolution move | It is fixed manually in the PR |
| **N6** | Stray `#define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING` with no `<codecvt>` include | `testOutlineManager.cpp:15`, kept because need the include indirectly |
| — | CI needs a green re-run on the head commit before merge (the recorded failure was an infrastructure cancellation at the dependency-install step, not a code failure) | HVTEx review §5, need to push changes and run the CI |

julia-b-grenier:
Alain provided additional suggestions generated by Claude after the last update. These suggestions have been addressed, and the following document summarizes the issues identified in each finding and the approach taken to resolve it.

# KBSH-1269 phase 2 — response to the HVT PR #186 review, Part 2

> **Started:** 2026-08-03 · **updated:** 2026-08-04 — all Part 2 items committed (`e0fe35f`..`db3f0f6`)
> **PR:** https://github.com/Autodesk/hydra-viewport-toolbox/pull/186 — "KBSH-1269 - Add the OutlineManager"
> **Review answered:** `KBSH-1269-OutlineManager-HVT-Part2.md` (re-reviewed at `142f1fd`) — findings **N1–N7** plus three Part 1 residuals
> **Part 1 response:** [review-response-phase2.md](./review-response-phase2.md) · **plan:** [review-part2-plan.md](./review-part2-plan.md)

Part 2 confirmed 10 of Part 1's 14 findings fully resolved and 2 mitigated to judgment calls. This
document answers what it raised on top of that. One row per item; the reasoning behind each conclusion is
spelled out below the tables, because in four places the conclusion differs from what the review proposed.

A final read of the range afterwards turned up two gaps in this pass's own work, both since fixed
(`98c5546`):

- **`outline_installSecondManagerOnSamePassIsRefused` did not pin N2.** It was written assuming a
  `TF_CODING_ERROR` fails an HVT test. It does not: `DiagnosticDelegate::IssueError`
  (`test/RenderingFramework/UsdHelpers.h:44-55`) only writes to `stderr`. With N2's pre-check
  reverted, every assertion still passed — `Install()` would set `framePass`, fail all five
  `AddTask` calls, and the destructor would then loop over five *empty* IDs and remove nothing,
  leaving the first manager's tasks intact. The test proved only an end state that already held
  before N2. Now a `TfErrorMark` around `Install()` asserts the error list stays clean, which is the
  actual discriminator: refused-before-`AddTask` posts nothing, `AddTask`-discovers posts five.
  `TF_WARN` cannot mask it — warnings go to the delegates only and are never appended to
  `TfDiagnosticMgr`'s per-thread `ErrorList` (`pxr/base/tf/diagnosticMgr.h:60`, `:188-190`, `:207-210`),
  which is the list a mark inspects.
- **`docs/outline.md`'s Unit Tests section** described `Install()` coverage but omitted the lifetime
  contract N1 and N2 introduced, which is the most consequential change in the pass.

Reviewing the flow-kitbashing docs against the same range then surfaced N1's real problem (see the
revert note under N1) plus three stale claims, all since fixed in
`docs/components/{OutlineProvider,OutlineRendering}.md`: the `Install()` row still said "destroying
the manager first is safe: the next commit no-ops via a weak-ptr guard" with no `~OutlineManager()`
row or one-per-pass constraint; `OutlineRendering.md` said `Execute()` "**also**" derives
`hasDistinct*`, implying a second setter; and both test-inventory rows said only "lifetime safety".
No app-side *code* change was needed — `AggregatorViewport` only calls `Install` / `SetStyle` /
`SetInputs`, and nothing in Falcon removes tasks.

## Status

| Item | Verdict | Status |
|---|---|---|
| Finding 11 residual — `docs/outline.md:19` ordering wording | Real, one word | **Fixed** `e0fe35f` |
| Finding 2 residual — are the regenerated baselines alpha-only? | **Confirmed alpha-only; no colour regression** | **Verified**, no code change |
| Finding 4 residual — can `_isStormRenderer` go stale? | **No; invariant holds** | **Verified**, no code change |
| PR description additions | Real | **Fixed** (`hvt.md`) |
| N1 — destroying the manager freezes the outline | Real symptom, **wrong fix** — the manager does not own the tasks | Fixed `83b0c51`, **reverted** `d8bd1d1`; documented instead |
| N2 — one-manager-per-frame-pass unstated | Real, and worse than filed: also a code defect | **Fixed** `2d01898` |
| N3 — two owners for `hasDistinct*` | Real; only one of the two directions is available | **Fixed** `97ea2eb` |
| N4 — `!_InitIfNeeded()` warn-latch | Real, but the proposed patch would not have fixed it | **Fixed** `7769c55` |
| N5 — `Prepare()` gate assumption unstated | Real — and the assumption turned out to be **provable** | **Fixed** `12e1867` |
| N6 — doc index / agent guide route to the demoted path | Real, at three sites | **Fixed** `12e1867` |
| N7a — missing `<cstdint>` | Real | **Fixed** `db3f0f6` |
| N7b — "unused" `PXR_NAMESPACE_USING_DIRECTIVE` | **Refuted** — it is load-bearing; the inverse fix applies | **Fixed** `db3f0f6` |
| N7c — namespace nesting + wrong close comment | Real | **Fixed** `db3f0f6` |
| N7d — `Immortal` tokens from mutable params | Real | **Fixed** `db3f0f6` |
| N7e — `=` misalignment in the new examples | Real | **Fixed** `db3f0f6` |
| Findings 9/M5, 12, 14 residuals | Judgment calls | **Not fixing** — rationale below |

---

## Part 1 residuals — closed

### Finding 11 residual · `docs/outline.md:19` · **fixed**

**What the review asked.** The intro paragraph said the manager owns "the fixed **layer** ordering". That
is a softer version of the phrase Part 1 flagged: what `Install()` owns is the task *execution* order; the
layer priority is a shader property that holds regardless of who installed the tasks.

**Fix.** `docs/outline.md:19` — "the fixed layer ordering" → "the fixed task ordering". The three
remaining `Overlay > Base > Default` uses (`:42`, `:54`, `:71`) sit in the layer-priority sections, where
the phrase is accurate, and are untouched.

### Finding 2 residual · the regenerated baselines · **verified alpha-only, no colour regression**

**What the review asked.** Only the *alpha* blend factors changed, so RGB should be bit-identical to the
pre-change baselines. It asked for a channel-wise diff of old vs. new, "otherwise a colour regression
rides in labelled 'expected baseline update'". Its circumstantial support was that the new files are all
smaller than the old, consistent with alpha becoming uniform.

**Two corrections to the premise, both of which change how the check has to be run.**

1. **Nine baselines were regenerated, not four.** All by `c6825ca` "Fix the outline tests", which is also
   the commit carrying the blend change (`source/tasks/outline/outlineOverlayTask.cpp`, alpha src factor
   `SrcAlpha` → `One`). The full set: `outline_renderEnabled`, `outline_renderBlurModes`,
   `outline_renderNonBasePrefix`, `outline_renderSelectedPath`,
   `outline_renderStyleChange_{none,blur3x3,blur5x5}`, `howTo/useOutlineManager`, `howTo/useOutlineTasks`.
2. **The "before" state is `c6825ca^`, not `d3c4b89` (#181).** Not every baseline was last touched in
   #181 — `outline_renderSelectedPath.png` does not exist in that tree at all
   (`git show d3c4b89:…` → "exists on disk, but not in 'd3c4b89'"). Comparing against the introducing
   commit would have silently skipped a file and mis-attributed the rest.

**Result — clean.** Channel-wise diff of all nine across `c6825ca^ → c6825ca`:

> R: identical · G: identical · B: identical · A: 60000/60000 px differ, max delta 255

**Why that is exactly the right shape**, not merely "no RGB delta": the defect was that the old
`(One, Zero)` alpha blend *overwrote* the destination alpha with the mask's alpha, which is ~0 everywhere
except near outline edges. So the pre-fix baselines should have near-zero alpha over the whole quad and
the post-fix ones should be opaque — a full-coverage change with delta 255. That is what the numbers show.
A partial alpha change, or any RGB movement, would have meant something else happened in that commit.

**One thing the check caught that would otherwise have produced a wrong claim.** Three of the nine were
regenerated *again* later, by `865ef1d`, so "current baselines vs pre-PR are alpha-only" is false for
them. Diffing `865ef1d^ → 865ef1d` isolates it:

| Commit | Baselines | R/G/B | A |
|---|---|---|---|
| `c6825ca` (blend change) | all 9 | **bit-identical** | 60000/60000 px differ, max delta 255 |
| `865ef1d` | the 3 `outline_renderStyleChange_*` | ~15000/60000 px differ, max delta 152–207 | **identical** |

The second row is deliberate and documented in that commit's own message — "camera 2x zoom added to
`outline_renderStyleChange` so blur modes are distinguishable" — i.e. a test-scene change, with alpha
untouched. The two effects are cleanly separated: the blend change never moved a colour channel, and the
colour change never moved alpha. Nothing else touched these nine after `c6825ca`
(`git log c6825ca..HEAD -- <the nine>` returns only `865ef1d`).

**Tooling** (kept in this folder, deliberately outside the repo): `png_channel_diff.py` — a stdlib-only
PNG decoder (`zlib` + `struct`, 8-bit non-interlaced), because neither Pillow nor numpy is declared in
`tools/pyproject.toml` and there was no image-diff script to reuse; and `verify_outline_baselines.py` —
extracts both blobs per file via `git show` and reports per-channel deltas. Re-runnable:
`python ".../phase2/verify_outline_baselines.py" [OLD_REV NEW_REV [paths…]]`.

### Finding 4 residual · `_isStormRenderer` staleness · **verified, invariant holds**

**What the review asked.** The `!_Enabled()` early return clears the dirty bits, which stops `_Sync`
running again, so `_isStormRenderer` is only refreshed if a renderer switch recreates the tasks. The
comment at `outlineMaskTask.cpp:576-584` asserts it does. Part 1 and Part 2 both left this as "plausible
but not verified — worth a one-line confirmation from an HVT owner".

**It is checkable in-repo, and it holds. Two independent links, both needed.**

1. **A renderer switch necessarily produces a new `HdRenderIndex`.** The render delegate is chosen and
   bound at construction — `renderIndexProxy.cpp:60-63`:
   `registry.CreateRenderDelegate(TfToken(rendererName), settingsMap)` followed by
   `HdRenderIndex::New(_renderDelegate.Get(), { hgiDriver })`. There is no setter to swap it afterwards:
   `grep 'SetRenderer|SetRenderDelegate'` over `include/hvt/engine` returns nothing. And
   `RenderIndexProxy` takes the renderer name in its constructor and is non-copyable
   (`renderIndexProxy.h:68`, `:70`), so changing renderer means constructing a new proxy.
2. **The tasks cannot outlive that index.** `TaskManager` receives `HdRenderIndex*` at construction
   (`taskManager.h:84`, `:89`), stores it (`:255`), and exposes no setter. `FramePass::Initialize` builds
   a fresh `TaskManager` from `frameDesc.renderIndex` (`framePass.cpp:221-222`) — so even the
   re-`Initialize` path replaces `_taskManager` wholesale, destroying its tasks with it.

Together: there is no route by which an `OutlineMaskTask` instance survives a change of render delegate.
`_isStormRenderer` is therefore fixed for the life of the task and cannot go stale, so clearing the dirty
bits on the non-Storm path is safe. **No code change; the existing comment is accurate as written**, and
the review's residual ask is satisfied without needing an owner's sign-off.

### PR description · **fixed** in `hvt.md`

The review could not check this from the diff. Five changes made:

1. **Removed an inverted claim.** The description still said "clear dirty bits when params fetch fails" —
   the opposite of what shipped (`71326246` leaves them set so a later re-sync retries; `0b1ff20` adds the
   latched warning). Replaced with the actual behaviour and the reason `SetTaskValue`'s equality gate makes
   the clear permanent.
2. **"4 updated image baselines" → 9**, and the alpha-only claim upgraded from asserted to measured.
3. **New `### Breaking changes` section** — both breaks, with no deprecation shim: the `active` → `lead`
   rename field by field, and the `SetVisualizationMode` removal *with why it had to go* (it wrote the
   task's local params copy and then marked `DirtyParams`, so the next `_Sync` refetched and overwrote it;
   a task has no route to the `TaskManager` that owns the stored params) plus the two paths that work.
4. **New bullet for the `Prepare()` resolution move**, which the description omitted entirely.
5. **Testing section** now carries the strongest evidence in the PR, which was previously buried in a
   baseline file nobody diffs: `test/data/baselines/howTo/useOutlineManager.png` and
   `useOutlineTasks.png` are the **same git blob** — confirmed at HEAD,
   `5bbe99a66a6d80f2ec6f5a342620f30133f6f7b1` for both (`git ls-tree HEAD …`). The wrapper is verified
   pixel-identical to the manual five-task wiring, not merely "comparable". Added alongside it the coverage
   caveat the review raised twice: every outline render test is `DISABLED_` on `__APPLE__`
   (`testOutlineManager.cpp:1383`, `:1487`, `:1608`, `:1776`) and `GTEST_SKIP`ed on Vulkan (`:1391`,
   `:1495`, `:1616`, `:1784`), so image-baseline coverage is desktop GL — the reason being
   non-deterministic `primId` rendering on Metal (`docs/outline.md:386-387`, `:406-407`).

Incidentally verified while patching: the description's existing claim that the new blend state matches
"the existing convention in `wboitResolveTask`" is accurate — `wboitResolveTask.cpp:139-141` is
`(SrcAlpha, OneMinusSrcAlpha)` for colour and `(One, OneMinusSrcAlpha)` for alpha.

---

## New findings — addressed

### N1 · Destroying the manager froze the outline on screen · **fixed** (`83b0c51`)

**Confirmed as filed, and it was the sharpest edge in the new API.** `~OutlineManager` was defaulted,
so it only dropped the last strong reference to the shared state. The five tasks stayed installed
*and enabled*, and every commit callback takes a `stateWeak.lock()`-failure path that returns
**before writing any params** — so `params.enabled` stayed `true` with the last pushed selection and
`CommitTaskValues` kept handing the tasks to `Engine::Execute`. The header sold the opposite: "no
explicit task removal is ever needed".

**Chose task removal over the documentation-only fallback**, on the strength of the one thing the
review had left unverified — whether `TaskManager::RemoveTask` is safe outside a render/commit phase.
It is: it erases from a `std::list` and calls the backend's `RemoveTask`; `~TaskManager` already
performs the same backend removal from a destructor; existing tests remove tasks with no commit in
between; and no phase or thread restriction is documented anywhere in `taskManager.h` — the class
holds no phase state that could enforce one. The single hazard is removing from *inside* a commit
callback, which would invalidate the iteration in `CommitTaskValues`; a destructor is not that.

Everything needed was already cached in `SharedState` (the `FramePass*` and the five `SdfPath` task
IDs), so this needed no new state.

**Three consequences shipped with it, all mandatory rather than cosmetic:**

1. The Install `\note` was inverted, not merely stale.
2. "Destroying the manager is safe from any thread" no longer holds — the destructor mutates the
   frame pass's task list, which the same note forbids off the render thread for `Install()`. The
   `weak_ptr` half of that claim still stands and is kept as the distinction: a racing *commit* is
   safe, the task-list mutation is not.
3. The `\pre` now requires the pass to outlive the manager's **destruction**, not merely its last
   commit — a genuinely stronger precondition, since a host violating it previously got a no-op
   destructor and now gets a dangling dereference.

#### N1 was then reverted (`d8bd1d1`) — the fix was wrong even though the symptom was real

Removing the tasks in the destructor treated the manager as owning them. It does not: it owns five
`SdfPath`s pointing into someone else's container. Three facts settle it, none of which I checked
before shipping `83b0c51`:

- `~TaskManager` already removes every task it holds (`taskManager.cpp:117-123`).
- `outlineManager.cpp` was the **only** `RemoveTask` call site in HVT outside `TaskManager`'s own
  implementation. Every other feature — the `createXTask` helpers, wboit, ssao, fxaa, visualizeAov —
  adds and never removes.
- No Falcon or NCC code removes tasks either: `grep RemoveTask` over `src/Extensions/`,
  `src/Core/Framework/`, `src/Apps/` returns nothing, and `AggregatorViewport` adds a dozen tasks
  through `renderTaskProxy` and removes none.

The header even said "The installed tasks are owned by the frame pass's TaskManager, not by this
manager" — and `83b0c51` kept that sentence while adding a destructor contradicting it.

**What it cost.** The destructor needed the pass alive, on the render thread, reached through a raw
pointer it cannot validate — a precondition Falcon's Qt path does not satisfy.
`NomadAdapter.cpp:361-363` builds the same `akUI::Aggregator::Viewport` that owns the manager, and
`~DDViewport` (`DDViewport.cpp:147-153`) queues the renderer's deletion onto the render thread and
then destroys `m_viewport` synchronously — so `RemoveTask` would run off the render thread. The web
path was safe by luck of declaration order (`WebViewport.h:167` before `:174`). And there is no
teardown counterpart to `setUpRenderer` in `akUI::Viewport`, so Falcon had no hook to do it properly.
Net for Falcon: an unreachable bug traded for a reachable one, since `m_outlineManager` and its pass
already die together.

**What shipped instead.** Defaulted destructor; the contract documents that destroying the manager
leaves the tasks installed with their last committed parameters, and that a host outliving its pass
pushes empty inputs first or calls `TaskManager::RemoveTask` itself. No `Uninstall()` was added —
`RemoveTask` is already the public way to do it.

Note the option the review offered as the lower-impact alternative (a final `enabled = false` commit)
carries the *same* thread requirement, since `SetTaskValue` also touches the pass. There is no way to
stop the outline without touching the pass, which is why this belongs to the host.

`outline_destroyManagerBeforeCommitIsSafe` keeps its original shape — the tasks survive and a later
commit no-ops — which also keeps the `stateWeak.lock()` failure path reachable and tested, instead of
the dead code `83b0c51` had turned it into. `outline_reinstallAfterRemovingTasksOnSameFramePass`
covers the documented route to reusing a live pass, and is the first HVT test to call
`TaskManager::RemoveTask(TfToken)` directly.

Left alone deliberately: `basePrimIdsTaskId` / `overlayPrimIdsTaskId` / `defaultPrimIdsTaskId` are
written and never read. Verified across the file's entire history that the destructor added in
`83b0c51` was their only reader ever — they were write-only from `4a750bc` onward. Kept for symmetry
with `maskTaskId` / `overlayTaskId`, which are read as insertion anchors.

### N2 · One manager per frame pass · **fixed** (`2d01898`) — and it was a code defect, not just a doc gap

**Worse than filed.** The review framed this as documentation. What actually happened on a collision:
all five `AddTask` calls fail with `TF_CODING_ERROR` and return empty paths, but `Install()` had
*already* cached the frame pass — so the second manager recorded itself as installed while owning
nothing, and its `SetInputs()` / `SetStyle()` then mutated state no task ever read. Five loud errors
followed by permanent silent no-ops.

So `Install()` now pre-checks the five fixed names via `TaskManager::HasTask` and returns before
writing any state: one actionable warning, the pass untouched, and the refused manager left
uninstalled — which also means its destructor has nothing to remove (the N1 guard covers it). The
`\pre` documents the constraint, and `docs/outline.md` records the refusal.

Also note this is why N1's removal matters beyond the visual bug: it frees the fixed names, so
destroy-then-reinstall on a live pass works instead of colliding.

### N3 · Two owners for `hasDistinctOverlay` / `hasDistinctDefault` · **fixed** (`97ea2eb`)

**Confirmed, but only one of the review's two "equally valid" resolutions is available.** Deleting
the recompute in `Execute` and keeping the manager as owner would regress every raw-task consumer:
`howTo20_UseOutlineTasks.cpp` deliberately never sets the flags and relies on `Execute` deriving them
from the texture names. So the manager's writes are the ones to go — and they were dead for rendering
anyway, since `Execute` recomputes unconditionally before the constants are pushed. They still took
part in `OutlineMaskStyleParams::operator==`, which is what made them a real second source of truth
rather than merely redundant. Note the same operator already compares all six texture names, so the
flags never added dirty-detection power the names did not already provide.

**The fallout was smaller than stated: 3 assertions in 2 tests, not 4 tests.** Two of the four
assertions the review listed compare against `0`, which is the `OutlineMaskStyleParams` constructor
default — so they passed regardless of what the manager pushed, i.e. they asserted nothing about it.
Those two were replaced with explicit base-alias comparisons, which is what they were actually
about; the three real ones moved onto the texture names.

**One doc error found beyond the finding:** `docs/outline.md` stated that the caller "points that
prefix's texture names at the `Base` names *and sets* `hasDistinctOverlay = 0`". Callers never set
it. Corrected, and the derivation is now recorded on the two public fields so a raw-task user does
not expect a pushed value to stick.

### N4 · `!_InitIfNeeded()` warn-latch · **fixed** (`7769c55`) — at a different site than proposed

**Right finding, wrong site: the review's patch would have added a message without silencing
anything.** The `if (!_InitIfNeeded())` block emits no diagnostic of its own. The per-frame
repetition comes from the two `TF_CODING_ERROR`s *inside* `_InitIfNeeded`, which re-fire because both
failure paths leave `_renderPass` / `_renderPassState` null while `_Sync` deliberately keeps its dirty
bits set to retry. A latched summary `TF_WARN` at the call site would have left those untouched.

So the latch went inside `_InitIfNeeded`, cleared on the success return so a failure that later
recovers reports again. **Also one file, not three:** `OutlineMaskTask::_InitIfNeeded` returns `void`
with no failure branch and is called unchecked; `OutlineOverlayTask` has no such function at all.

Left alone deliberately, and stated in the commit: `_CreateAovBindings()` has the same shape — it is
re-entered while `_aovBuffers` stays empty and raises three unlatched coding errors, as does
`OutlineMaskTask`. Neither reports failure through a return value, so they need a different fix.

No test. Reaching this path requires `CreateRenderPass` or `_InitIdRenderPassState` to fail, which is
not injectable here, and the assertion would be on diagnostic frequency rather than outline
behaviour.

### N5 · The `Prepare()` gate's assumption · **fixed** (`12e1867`) — and the assumption is provable

The review asked for the invariant to be written down, as Part 1's finding 4 was. Recording it as an
assumption would have been the easy answer; it is instead **provable from the OpenUSD source**, so
the doc now states a fact with citations rather than a caveat.

Verified against **v25.11 and v26.05, identical on every point**:

| Link | Evidence |
|---|---|
| `HdRprim::SetPrimId` is the only writer of `_primId`, and only stores it | `hd/rprim.cpp` — body is `_primId = primId;` plus "Don't set DirtyPrimID here, to avoid undesired variability tracking" |
| It is reached from exactly two places | `_AllocatePrimId` and `_CompactPrimIds`, both `hd/renderIndex.cpp` |
| `_AllocatePrimId` is called only from `_InsertRprim` | `hd/renderIndex.cpp` |
| `_CompactPrimIds` is called only from `_AllocatePrimId` | so even the 24-bit-exhaustion reshuffle of *every* id is part of an insertion |
| `_InsertRprim` calls `HdChangeTracker::RprimInserted`, which does `++_rprimIndexVersion` | `hd/changeTracker.cpp`; `RprimRemoved` bumps it too |

**The subtlety worth recording** — and the reason the gate is sound for a slightly different reason
than the H1 decision record originally assumed: `_CompactPrimIds` marks every rprim `DirtyPrimID`,
but `_MarkRprimDirty` increments `_rprimIndexVersion` only for `DirtyRenderTag` / `DirtyRepr`. So the
guarantee comes from the enclosing insertion, not from that dirty bit. `code-review-phase2.md` has
been updated with the same citations.

Scope checked: `hd/renderIndex.cpp`, `hd/rprim.cpp`, `hd/changeTracker.cpp` at both tags.
`SetPrimId` is public `HD_API`, so a caller elsewhere in `pxr` or in a third-party render delegate is
not excluded by the API — but none of those three files calls it, and the scene-index emulation path
routes through `AddLegacyPrim` to the same `_InsertRprim`.

### N6 · Doc index and agent guide · **fixed** (`12e1867`) — three sites, one more than filed

`docs/README.md`'s Features row and `AGENTS.md`'s task row both routed readers to the raw-task path
alone, so anyone following them lands on `howTo20` and rediscovers the wiring by hand. The review
named those two; `docs/README.md` also carries a separate how-to list with a `howTo20` row and no
`howTo21` counterpart. All three now name both, marking which to reach for first. Neither file was in
the branch diff before this.

### N7 · Nits · **fixed** (`db3f0f6`) — three real, one refuted, one reframed

**7a — missing `<cstdint>`.** Real; `uint64_t` worked only transitively. Added.

**7b — "unused" `PXR_NAMESPACE_USING_DIRECTIVE`. Refuted, and the correct fix is the inverse of the
one proposed.** The directive is **load-bearing**: `TF_WARN` expands to unqualified
`Tf_PostWarningHelper` / `TfCallContext`, which resolve only through it, so deleting it breaks the
build. The review read it as unused because every *explicit* pxr reference in the file is
`PXR_NS::`-qualified — and that is the actual inconsistency: `grep -c 'PXR_NS::'` over
`source/tasks/` gives **26 for `outlineManager.cpp` and 0 for every other `.cpp`**, including its
three outline siblings. So the directive stays and the 33 qualifications go. The header keeps
`PXR_NS::`, since headers carry no directive and the siblings qualify there too.

**7c — namespace nesting and a wrong close comment.** Real. Rather than fix only the comment, the
forward declaration moved into its own `namespace HVT_NS { ... }` block so the file can use the
flattened `namespace HVT_NS::Outline` its four siblings use — which removes the blocker the review
identified *and* makes the existing close comment correct.

**7d — `Immortal` tokens from mutable params.** Real. The members own the tokens for the task's
lifetime, so `Immortal` bought nothing and pinned one registry entry per prefix ever seen. Dropped,
with a note that the genuinely constant names in `outlineTextureNames.h` are correctly `Immortal`.

*Adjacent, outside the finding:* `OutlinePrimIdsTask::GetToken` also interned `Immortal`, into a
function-local static map. That one was redundant rather than leaky — the map owns every token for
the life of the process, so `Immortal` reclaimed nothing either way and the map, not `Immortal`, is
what grows per distinct prefix. Dropped for consistency with the rule this pass establishes
(`Immortal` marks a genuine constant), with no behaviour change. Returning `TfToken` by value and
deleting the map plus its mutex would remove the growth as well, but that changes a public signature
for a nit and the only production caller passes the three fixed prefixes — logged rather than done.

**7e — `=` misalignment. Reframed: a 3-line hand fix, not a formatting pass.**
`.clang-format` sets `AlignConsecutiveAssignments: true`, so the target alignment is reachable by
editing exactly three lines. Running clang-format over the whole file would have widened the diff
past the change.

The line wraps were left to clang-format rather than hand-reflowed: dropping the qualifications
shortens lines that were wrapped for the longer names, and predicting the tool's choice
(`BinPackParameters` vs `AllowAllParametersOfDeclarationOnNextLine: true`) is guesswork.

---

## Where this response diverges from the review

Collected for the reviewer's convenience — four places where the finding is real but the proposed fix is
not what should ship, and one refutation.

| Item | Divergence |
|---|---|
| **N7b** | **Refuted.** `PXR_NAMESPACE_USING_DIRECTIVE;` (`outlineManager.cpp:42`) is not unused — `TF_WARN` at `:146` expands to unqualified `Tf_PostWarningHelper` / `TfCallContext` (`pxr/base/tf/diagnostic.h:371-372`, `pxr/base/tf/callContext.h:30-31`), so deleting it breaks the build. The review read it as unused because every *explicit* pxr reference in the file is `PXR_NS::`-qualified — and that is the real inconsistency: `grep -c 'PXR_NS::'` gives **26 for `outlineManager.cpp` and 0 for every other `.cpp` under `source/tasks/`**, including its three outline siblings. So the fix is the inverse of the one proposed: keep the directive, drop the 33 qualifications. The header keeps `PXR_NS::` — headers carry no using-directive, and the siblings qualify there too. |
| **N4** | **Right finding, wrong site.** The `if (!_InitIfNeeded())` block (`outlinePrimIdsTask.cpp:335-343`) emits no diagnostic at all; the per-frame repetition comes from the two `TF_CODING_ERROR`s *inside* `_InitIfNeeded` (`:147`, `:154`), which re-fire because both leave `_renderPass`/`_renderPassState` null. A latched summary `TF_WARN` at the call site, as proposed, adds a message without stopping the spam — latch inside `_InitIfNeeded` instead. Also this is one file, not three: `OutlineMaskTask::_InitIfNeeded` returns `void` with no failure branch (`outlineMaskTask.cpp:180-194`, called unchecked at `:622`), and `OutlineOverlayTask` has no such function. |
| **N3** | **Only one of the two "equally valid" directions is available.** Deleting the recompute in `Execute` and keeping the manager as owner would regress every raw-task consumer: `howTo20_UseOutlineTasks.cpp:202-206` deliberately never sets the flags and relies on `Execute` deriving them from the texture names. So the manager's writes are the ones to delete. The fallout is also smaller than stated — 3 assertions in 2 tests (`testOutlineManager.cpp:641`, `:645`, `:1370`), not 4 tests; `:561` and `:583` assert `== 0`, already the constructor default (`outlineMaskTask.h:95-96`), so they pass either way. |
| **N2** | **Worse than filed** — not merely undocumented. Two managers on one pass make all five `AddTask` calls fail with `TF_CODING_ERROR` and return empty paths (`taskManager.cpp:171-180`), yet `Install()` has already set `_state->framePass` (`outlineManager.cpp:161`), so the second manager believes it is installed and its `SetInputs`/`SetStyle` mutate state nothing reads. A silent no-op behind a loud error. |
| **N1** | **Symptom confirmed, fix withdrawn.** The frozen outline is real, but removing the tasks in `~OutlineManager` was wrong: `TaskManager` owns them and destroys them itself, and no other feature in HVT or the app removes a task. Shipped as documentation plus the existing `TaskManager::RemoveTask` escape hatch. Note the review's fallback (a final `enabled = false` commit) carries the same thread requirement, so no destructor-side option was available. |
| **N5** | **Asked for an assumption to be recorded; it is provable instead.** Following `SetPrimId` -> `_AllocatePrimId` / `_CompactPrimIds` -> `_InsertRprim` -> `RprimInserted` in OpenUSD v25.11 and v26.05 shows every primitive-ID assignment happens inside an insertion, which bumps the version. The doc states that with citations rather than hedging. It also corrects the reason: the bump comes from the enclosing insertion, not from `_CompactPrimIds`' own `DirtyPrimID` mark, which `_MarkRprimDirty` ignores for versioning. |
| **N7e** | **Reframed as a 3-line hand fix, not a formatting pass.** `.clang-format:9` sets `AlignConsecutiveAssignments: true`, so the target alignment is reachable by editing exactly three lines (`howTo21_UseOutlineManager.cpp:165`, `:184`, `docs/outline.md:281`). Running clang-format over the whole file would contaminate the diff. |

---

## Deliberately not fixing

| Item | Reason |
|---|---|
| Finding 9 / M5 — `CacheStats` in a public header | Accepted as a judgment call on API surface, which is where Part 2 landed it too ("now a judgment call rather than a correctness or clarity problem. Not a merge blocker"). The semantics are fully documented on `GetCacheStats()` and pinned by two tests, and `collectionSizeSum` — the misleadingly named field — is not public; it is a private `SharedState` member. |
| Finding 12 — the `TF_DEBUG_ENVIRONMENT_SYMBOL` reformat | Kept. Part 2 concedes the reason ("consistency across the outline files is a defensible reason to keep it"), and it was applied deliberately across both outline task sources rather than incidentally. Reverting it now would itself be churn. Called out in the PR description instead. |
| Finding 14 residual — no test for the raw-task `maskVisualizationMode` push | The substantive leg — that the params value actually selects a compute program — is covered by `outline_renderVisualizationModes` (four per-mode GPU baselines). A raw-task round-trip would only re-assert that `TaskManager::SetTaskValue` stores what it is given, which is not outline behaviour. Part 2 rates it "low value; optional" and we agree. |

laferra-adsk (approved):
Awesome work! Thanks for your diligence with all the off-PR claude analysis iterations and improvements.

julia-b-grenier:
Review rounds 3 and 4 — summary

Since the last review pass the branch has taken nine commits answering two review rounds, d8bd1d1 → 12330f9. No behavioural defect was found in either round; everything below is either a fix to something a comment or doc asserted incorrectly, or one genuine coverage gap.

Round 3 (5 commits) — four findings, plus twelve further corrections from re-reading each fix against the code rather than trusting the commit message:

- The documented teardown recipe didn't work. It now requires cleared inputs and a style with enableDefaultOutlines = false and one commit, and says why each is needed.
- OutlinePrimIdsTask could report successful initialization with a null render-pass state, and a partially-built AOV binding set could survive for the life of the task. Both closed; _CreateAovBindings() now reports failure instead of it being inferred.
- Prim-ID resolution moved to Prepare() in an earlier commit, but the manager still credited _Sync() and claimed it ran every frame. Corrected, along with three inverted lifetime sentences and an unreachable branch that had been documented rather than deleted.

Round 4 (4 commits) — four findings, all in prose except one:

- docs/outline.md said outline_renderLeadPicksUpInsertedPrim "needs no baseline". It has one, added by the same commit, and the test asserts against it — effectively written permission to delete a baseline the suite depends on. Fixed.
- The cleared-inputs teardown route had no test, and the case previously cited as covering it makes two pushes that are both dedup no-ops on a fresh manager, so it asserted the initial state and never a transition. outline_clearedInputsAndStyleDisableEveryTask now pins the full ordering, including that clearing inputs alone leaves the tasks enabled while enableDefaultOutlines is set.
- The GPU-case enumeration counted three of four, omitting outline_renderSelectedPath; the how-to still carried a lifetime rationale retired a round earlier; the task source/header lists are now sorted.

Two things deliberately not changed, with reasoning rather than silence:

- The AOV dimension-check diagnostic repeats every sync while the render-pass ones are latched. Reaching it means the host committed an unsized frame pass — a contract violation, not a recoverable state — so the repeating error is the intended pressure; latching it would make a programming bug quieter.
- The remaining initialization-failure paths stay untested. There is no failure-injection seam in the suite, and the only cheaply reachable route is that same contract violation. Every path reachable under Storm is already covered by the render cases.

Here are the details of each round of review: 
[review-response-part3.md](https://github.com/user-attachments/files/30803237/review-response-part3.md)
[review-response-part4.md](https://github.com/user-attachments/files/30803239/review-response-part4.md)

hodoulp (approved):
LGTM

PR #192, opened 2026-08-07
AGP-670 - Fix code-review issues and add regression tests

hodoulp:
## Description

Addresses a batch of defects surfaced during a full code review of the Hydra Viewport Toolbox, grouped by severity, plus some targeted unit tests. No public API changes: every fix is either defensive (bounds checks, null guards) or a correctness fix to existing behavior.

### Changes Made

**Critical / High**

- `flashPickTask`: clamp the pick sub-region to the render buffer to stop an out-of-bounds read when the pick frustum extends past the buffer edges.
- `pageableDataSource`: reuse a page slot only when its size matches the packed blob; otherwise reallocate, avoiding OOB reads / on-disk truncation.
- `renderBufferManager`: guard a null color backing buffer before dereferencing.
- `geometry`: bail out when the 2D material surface shader fails to parse instead of dereferencing a null `Sdr` node.

**Medium**

- `renderBufferManager`: decide the progressive-render clear from the real previous-vs-new AOV output comparison.
- `lightingManager`: remove every excess light when the active count shrinks by more than one.
- `boundingBoxSceneIndex`: announce the same prim type `GetPrim` converts to, and remap `extent` invalidation to also dirty `primvars:points`.
- `composeTask`: null-guard the `ComposeTaskParams` stream operator.
- `fxaaTask`: derive `uResolution` from the color texture so it tracks viewport resizes.

**Low**

- `copyDepthShader`: destroy shader functions on the compile-failure paths.
- `taskManager`: guard `++end()` on `insertAfter` and skip null tasks from the render index.
- `viewportEngine`: bounds-check `xformOps` before indexing.
- `ssaoTask`: clamp `sampleCount` to avoid OOB table access and divide-by-zero.
- `wboitRenderTask`: drop a dead early-out that never fired.
- `dataSourceRegistry`: bounds-check `getFileTypesDesc(index)`.
- `shadowMatrixComputation`: guard degenerate (zero) light positions.
- `dataSource`: default-initialize `AreaLightInfo::type`.
- `selectionHelper`: null-check the `engine` argument.

**Tests**

- New GPU-free scene-index test (`testBoundingBoxSceneIndex.cpp`) for `BoundingBoxSceneIndex` notice plumbing: the announced prim type matches `GetPrim`, non-convertible prims pass through unchanged, and dirtying `extent` also dirties `primvars:points`.
- Regression tests for the registry bounds check, `AreaLightInfo` default type, `TaskManager` `insertAfter`-at-end, and the `ComposeTaskParams` null stream operator.

## Testing

Built with warnings-as-errors and ran the newly added unit tests locally. A few fixes (progressive-clear, FXAA resolution, shader-function leak, lighting reconciliation, pageable re-page-on-resize) are not reachable through public APIs without a GPU pipeline or extra harness, so they are covered indirectly by existing image-comparison tests rather than new unit tests.

### Test Configuration
- **Platform(s) tested**: macOS 15 (Darwin 25.5.0, Apple M3 Pro)
- **Build configuration(s)**: All cmake presets
- **OpenUSD version**: 0.26.5 (v26.05)

### Tests Performed

## Documentation

## Checklist

### DIFF
--- a/source/tasks/flashPickTask.cpp
+++ b/source/tasks/flashPickTask.cpp
@@ -366,6 +366,16 @@ void FlashPickTask::Execute(HdTaskContext* ctx)
     GfVec4i subRect(static_cast<int>(pickMin[0]), static_cast<int>(pickMin[1]),
         static_cast<int>(pickMax[0] - pickMin[0]), static_cast<int>(pickMax[1] - pickMin[1]));
 
+    // Clamp the sub-region to the render buffer. The projected pick frustum can extend
+    // outside the buffer (cursor near an edge, or a pick camera that differs from the
+    // render camera); without this the loops below would index the mapped ID/depth
+    // buffers (which hold exactly width*height elements) out of bounds.
+    const int minX = std::max(0, subRect[0]);
+    const int minY = std::max(0, subRect[1]);
+    const int maxX = std::min(width, subRect[0] + subRect[2]);
+    const int maxY = std::min(height, subRect[1] + subRect[3]);
+    subRect        = GfVec4i(minX, minY, std::max(0, maxX - minX), std::max(0, maxY - minY));
+
     // Resolve hits using Flash's primId-to-path map (bypasses HdRenderIndex).
     // Resolve code copied from HdxPickResult.
     HdRenderDelegate* rd = _index->GetRenderDelegate();
--- a/include/hvt/dataSource/dataSource.h
+++ b/include/hvt/dataSource/dataSource.h
@@ -207,7 +207,9 @@ public:
     struct AreaLightInfo
     {
         PXR_NS::SdfPath path;
-        AreaLightType   type;
+        // Default-initialized so a caller that reads `type` after a failed getAreaLight()
+        // (which returns false without touching the struct) does not read an indeterminate value.
+        AreaLightType   type = AreaLightType::Rect;
     };
 
     /// \brief Creates a new area light prim of the specified type.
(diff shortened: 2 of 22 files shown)


source/tasks/flashPickTask.cpp
```diff
@@ -366,6 +366,16 @@ void FlashPickTask::Execute(HdTaskContext* ctx)
...
         static_cast<int>(pickMax[0] - pickMin[0]), static_cast<int>(pickMax[1] - pickMin[1]));
 
+    // Clamp the sub-region to the render buffer. The projected pick frustum can extend
+    // outside the buffer (cursor near an edge, or a pick camera that differs from the
+    // render camera); without this the loops below would index the mapped ID/depth
+    // buffers (which hold exactly width*height elements) out of bounds.
+    const int minX = std::max(0, subRect[0]);
+    const int minY = std::max(0, subRect[1]);
+    const int maxX = std::min(width, subRect[0] + subRect[2]);
+    const int maxY = std::min(height, subRect[1] + subRect[3]);
```
sebastienberube-adsk, line 376: Perhaps this is fine, but wondering if this should be clamped to `width-1` and `height-1`.
(is this max inclusive [min, max] or exclusive [min, max[?)

sebastienberube-adsk (approved):

source/tasks/flashPickTask.cpp
hodoulp: That's a computation on a sub-region where the region is [0, width[.

PR #196, opened 2026-08-19
AGP-701 - Rendering test framework in one CMake module

lcpoulin-adsk:
## Description

Create a reusable cmake module for the RenderingFramework test helpers.

## Documentation

## Checklist

### DIFF
--- a/cmake/HvtTestFramework.cmake
+++ b/cmake/HvtTestFramework.cmake
@@ -1,0 +1,153 @@
+# Copyright 2026 Autodesk, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


cmake/HvtTestFramework.cmake
hodoulp, line 1: Can you explain why do you need this file?
The cmake code to create a library for the `RenderingFramework` can be in the `test/RenderingFramework/CMakeList.txt` and will never be reused by some else.

hodoulp:
Note: As this library will be used other libraries the install step must install include, binaries and cmake generated files. 

To simplify the library creation (and because it's helper for unit tests) I mentioned that you can hard-code it to be a static library. If you do not impose that you will have to manage the dynamic library case (with all the mess around import/export symbols on Windows).

PR #198, opened 2026-08-28
AGP-786 - Speed up the build ~2.7x

hodoulp:
## Description

The main goal of the pull request is to speed up the build.

### How To Test

The command below provides a way to manually test the build cost on your machine:
`start=$SECONDS; ninja all; end=$SECONDS; echo "time => $(($end-$start))"`

### Results

Clean builds, 12-core Apple M3 Pro, `main` (8548e54) vs this branch. All measurements taken back-to-back in a single session with the compiler cache disabled, so they are comparable to each other:

| Configuration | | main | branch | speedup |
|---|---|---:|---:|---:|
| **Debug** | core library | 39s | 16s | **2.4x** |
| | tests | 44s | 15s | **2.9x** |
| | **complete** | **83s** | **31s** | **2.7x** |
| **Release** | core library | 42s | 14s | **3.0x** |
| | tests | 45s | 16s | **2.8x** |
| | **complete** | **87s** | **30s** | **2.9x** |
| **RelWithDebInfo** | core library | 42s | 16s | **2.6x** |
| | tests | 46s | 17s | **2.7x** |
| | **complete** | **88s** | **33s** | **2.7x** |

### Important Changes Made

- **Precompiled headers** (`ENABLE_PRECOMPILED_HEADERS`, default `ON`). `source/pch.h` covers the library, `test/pch.h` the tests. Both hold third-party headers only — an HVT header in a PCH would rebuild everything whenever it changed.
- **One PCH per flag group, not per target.** `hvt_enable_precompiled_header(<target> <header> <group>)` makes the first target in a group compile the header and the rest `REUSE_FROM` it.
  Two groups are needed because the tests do not compile with `HVT_BUILD` and do add `GLEW_STATIC`, and a PCH can only be reused where the definitions match. By default the ccache cache is global (for all builds whatever is the workspace) so, note that the cache is per repo to avoid potential conflicts with an arbitrary workspace.
- **Trimmed the most expensive public includes.**
- **Compiler cache** (`ENABLE_COMPILER_CACHE`, default `OFF`), auto-detected, skipped when HVT is
  a sub-project since `CMAKE_CXX_COMPILER_LAUNCHER` would apply to the whole build.
- **Limited debug info in the optimized builds** (`ENABLE_LIMITED_DEBUG_INFO`, default `ON`,
  `RelWithDebInfo` only). Apple's Clang defaults to standalone debug info, emitting a full
  definition for every type a TU mentions even when another TU already emits it.

    | Preset | Build | Link | Objects |
    |---|---|---|---|
    | `relwithdebinfo` | 31s → 27s | 3.80s → 3.04s | 405 MB → 133 MB |
    | `asan` | 33s → 30s | 3.65s → 2.72s | 447 MB → 182 MB |
    | `ubsan` | 34s → 32s | 3.61s → 3.08s | 406 MB → 133 MB |

### Test Configuration
- **Platform(s) tested**: macOS 15 (Darwin 25.5.0), Apple M3 Pro, 12 cores
- **Build configuration(s)**: all 8 presets — `debug`, `debugstatic`, `release`,
  `relwithdebinfo`, `asan`, `ubsan`, `debugwithvulkan`, `releasewithvulkan`
- **OpenUSD version**: 0.26.8 (vcpkg)

### Tests Performed

**Not tested:** Windows and Linux. The Windows ccache branch and the MSVC paths are reasoned about, not measured — worth a look from someone on those platforms.

## Documentation

## Checklist

- [x] **I have signed the Contributor License Agreement (CLA)**


cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+# Usage:
+#   hvt_setup_compiler_cache()
+#
+# NOTE: A compiler cache does nothing for a clean build; it pays off when rebuilding after
+# switching branches or reverting a change, where it turns a recompile into a cache hit.
+# NOTE: A launcher already set by the caller (a parent project, or -DCMAKE_CXX_COMPILER_LAUNCHER
+# on the command line) is left alone, but still gets the precompiled header adjustment below.
+# NOTE: Nothing is detected when HVT is built as a sub-project. Whether to use a compiler cache is
+# the top-level project's decision, and CMAKE_CXX_COMPILER_LAUNCHER would apply to all of it.
+function(hvt_setup_compiler_cache)
```
DDoS, line 15: The performance gain from the compiler cache are dubious, not worth the maintenance.

DDoS (commented):
Please clean up the excessive AI verbosity and comments before asking for a review. This also includes the PR description.

adskWangl:
I'd also prefer a clean build. Unless we have a stable prebuild pkgs, even the dependency cache is always causing quite a lot issues when doing the upgrade.

cmake/HvtHelpers.cmake
hodoulp: When localy compiling the library it has a huge impact, and I contacted Laurence to see if `ccache` can be on CI/CD build machines. Stay tuned

hodoulp:
> I'd also prefer a clean build. Unless we have a stable prebuild pkgs, even the dependency cache is always causing quite a lot issues when doing the upgrade.

All of that is to speed up the local build (which indirectly speeds up the CI/CD builds). However, the CI/CD build speed up will require much more work because most of the time is lost 'outside' the library build itself. The dependency is one of the critical point but not the only one.

hodoulp:
> Please clean up the excessive AI verbosity and comments before asking for a review. This also includes the PR description.

Yes, this is a recurring issue with AI i.e., too verbose but still contains important information for AI agent. I don't have a clear answer on this topic. I will investigate these changes. Stay tuned.

DDoS (changes requested):
This is way too complicated for such a small project. Reading the ccache comments gives me zero confidence it's actually worth it.
Also I've used PCH with CMake a lot, and it really really should just be one command:
```
target_precompile_headers(myTarget PRIVATE pch.h)
```
I don't understand where the complexity comes from in this PR, and I don't think it's justified. HVT is such a small project anyway, do we need to put this much effort in making it build faster and sacrifice more build legibility?

hodoulp:
> This is way too complicated for such a small project. Reading the ccache comments gives me zero confidence it's actually worth it.

That's exactly the point. **Why does such a small project take so long to build, especially on high-end machines?**
(e.g., M3 Pro, 36G)

Note: This then propagates to `HVTEx` and `Components`, impacting the hundreds of builds we have to run when fixing something in one of the VT repos.

On my machine, when switching between branches, rebuilding something, etc, the `ccache` provides a major build speedup.

DDoS:
> > This is way too complicated for such a small project. Reading the ccache comments gives me zero confidence it's actually worth it.
> 
> That's exactly the point. **Why does such a small project take so long to build, especially on high-end machines?**
> 
> Note: This then propagates to `HVTEx` and `Components`, impacting the hundreds of builds we have to run when fixing something in one of the VT repos.
> 
> On my machine, when switching between branches, rebuilding something, etc, the `ccache` provides gives a major build speedup.

Can you just set it up purely locally using a preset? I don't actually see why this needs to be in the CMake script. Ultimately all it seems to do is set `CMAKE_CXX_COMPILER_LAUNCHER` and some `CCACHE` environment variables.

I don't want us to go down the same road of increasing build complexity until we end up with the nightmare that is Component. This is really looking like over-engineered Claude nonsense. And if there's someone to say no to this, it's going to be me.

hodoulp:
> I don't want us to go down the same road of increasing build complexity until we end up with the nightmare that is Component. 

Note: That's an unfortunate historical legacy i.e., I suspect project was created as a POC using copy & paste from an existing one (could be an OGS project based on various names) without revisiting it at that time so we are now screwed with that and `DependencyTool` doing everything syndrom. That's not a degradation over time.

hodoulp:
> Can you just set it up purely locally using a preset? I don't actually see why this needs to be in the CMake script. Ultimately all it seems to do is set CMAKE_CXX_COMPILER_LAUNCHER and some CCACHE environment variables.

That's more than the presence or not of the tool and it's clearly encapsulated in a cmake helper. So, I think it's pertinent.

DDoS:
> > Can you just set it up purely locally using a preset? I don't actually see why this needs to be in the CMake script. Ultimately all it seems to do is set CMAKE_CXX_COMPILER_LAUNCHER and some CCACHE environment variables.
> 
> That's more than the presence or not of the tool and it's clearly encapsulated in a cmake helper. So, I think it's pertinent.

I really want to avoid adding more code to the CMake script, so it's doable through external configuration, that's how it should be done. Please give it a try.

hodoulp:
> > > Can you just set it up purely locally using a preset? I don't actually see why this needs to be in the CMake script. Ultimately all it seems to do is set CMAKE_CXX_COMPILER_LAUNCHER and some CCACHE environment variables.
> > 
> > 
> > That's more than the presence or not of the tool and it's clearly encapsulated in a cmake helper. So, I think it's pertinent.
> 
> I really want to avoid adding more code to the CMake script, so it's doable through external configuration, that's how it should be done. Please give it a try.

The fully benefit from `ccache`, you need `env CCACHE_DEPEND=1 CCACHE_SLOPPINESS=pch_defines,time_macros` and `add_compile_options("SHELL:-Xclang -fno-pch-timestamp")` so that PCH does not defeat the cache. That clearly demonstrates the need of an helper which is also used by `HVTEx`.

DDoS:
> > > > Can you just set it up purely locally using a preset? I don't actually see why this needs to be in the CMake script. Ultimately all it seems to do is set CMAKE_CXX_COMPILER_LAUNCHER and some CCACHE environment variables.
> > > 
> > > 
> > > That's more than the presence or not of the tool and it's clearly encapsulated in a cmake helper. So, I think it's pertinent.
> > 
> > 
> > I really want to avoid adding more code to the CMake script, so it's doable through external configuration, that's how it should be done. Please give it a try.
> 
> The fully benefit from `ccache`, you need `env CCACHE_DEPEND=1 CCACHE_SLOPPINESS=pch_defines,time_macros` and `add_compile_options("SHELL:-Xclang -fno-pch-timestamp")` so that PCH does not defeat the cache. That clearly demonstrates the need of an helper which is also used by `HVTEx`.

Nothing that can't be done in a preset. It's strictly not needed so this only needs to be README information for those who want to use it.
```
{
  "name": "base",
  "hidden": true,
  "generator": "Ninja",
  "binaryDir": "${sourceDir}/build/${presetName}",
  "environment": {
    "CCACHE_DEPEND": "1",
    "CCACHE_SLOPPINESS": "pch_defines,time_macros"
  },
  "cacheVariables": {
    "ENABLE_TESTS": "ON",
    "ENABLE_BENCHMARKS": "OFF",
    "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/${presetName}",
    "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
    "CMAKE_C_FLAGS": "-Xclang -fno-pch-timestamp",
    "CMAKE_CXX_FLAGS": "-Xclang -fno-pch-timestamp"
  }
}
```

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+#
+# Usage:
+#   hvt_setup_compiler_cache()
+#
+# NOTE: A compiler cache does nothing for a clean build; it pays off when rebuilding after
+# switching branches or reverting a change, where it turns a recompile into a cache hit.
+# NOTE: A launcher already set by the caller (a parent project, or -DCMAKE_CXX_COMPILER_LAUNCHER
+# on the command line) is left alone, but still gets the precompiled header adjustment below.
+# NOTE: Nothing is detected when HVT is built as a sub-project. Whether to use a compiler cache is
+# the top-level project's decision, and CMAKE_CXX_COMPILER_LAUNCHER would apply to all of it.
```
DDoS: Useless comment.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+        if (HVT_COMPILER_CACHE MATCHES "ccache")
+            # ccache's default mode is actively harmful here, and leaving it to each developer's
+            # global configuration is a trap, so the settings travel with the build.
+            #
+            # By default ccache runs the preprocessor to compute a cache key. For HVT that costs
+            # about as much as the compile, and with a precompiled header it costs far more: a
+            # clean build measured 34s with no compiler cache and 88s with a default-configured
+            # ccache. Depend mode keys off the compiler's own dependency output instead, which
+            # brings a cold-cache clean build back to ~36s and a warm one down to ~4s. The
+            # sloppiness settings are what let a precompiled header be cached at all.
```
DDoS: This comment could be one line.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+            if (NOT WIN32)
+                set(_launcher
+                    env CCACHE_DEPEND=1 CCACHE_SLOPPINESS=pch_defines,time_macros
```
DDoS: Why is `env` needed?

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+                    "${HVT_COMPILER_CACHE}")
+            else()
+                # env(1) is unavailable, so the settings cannot be injected per build. Enabling an
```
DDoS: Windows has a similar syntax: `set MY_VAR=HelloWorld & mycommand.exe`

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+                        "'ccache --set-config depend_mode=true' and "
+                        "'ccache --set-config sloppiness=pch_defines,time_macros'.")
+                    return()
+                endif()
+            endif()
+        endif()
+
+        # Set in the caller's directory scope rather than the cache, so that the setting is
+        # inherited by the sub-directories without being written to a parent project's cache.
+        set(CMAKE_CXX_COMPILER_LAUNCHER "${_launcher}" PARENT_SCOPE)
```
DDoS, line 47: Instead of injecting the environment variables in the command line, you could just set them on the current process. Even simpler.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+                        "Compiler cache disabled: ccache needs depend mode, otherwise a clean "
+                        "build is slower with it than without. Enable it once with "
+                        "'ccache --set-config depend_mode=true' and "
+                        "'ccache --set-config sloppiness=pch_defines,time_macros'.")
+                    return()
+                endif()
+            endif()
+        endif()
+
+        # Set in the caller's directory scope rather than the cache, so that the setting is
```
DDoS, line 45: This is intended to be used as a cache variable, why are we not using it as one? The comment doesn't actually explain.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+# on the command line) is left alone, but still gets the precompiled header adjustment below.
+# NOTE: Nothing is detected when HVT is built as a sub-project. Whether to use a compiler cache is
+# the top-level project's decision, and CMAKE_CXX_COMPILER_LAUNCHER would apply to all of it.
+function(hvt_setup_compiler_cache)
+    if (NOT ENABLE_COMPILER_CACHE)
+        return()
+    endif()
+
+    if (NOT CMAKE_CXX_COMPILER_LAUNCHER)
+        if (NOT PROJECT_IS_TOP_LEVEL)
```
DDoS: This check doesn't belong here.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+#   hvt_setup_compiler_cache()
+#
+# NOTE: A compiler cache does nothing for a clean build; it pays off when rebuilding after
+# switching branches or reverting a change, where it turns a recompile into a cache hit.
+# NOTE: A launcher already set by the caller (a parent project, or -DCMAKE_CXX_COMPILER_LAUNCHER
+# on the command line) is left alone, but still gets the precompiled header adjustment below.
+# NOTE: Nothing is detected when HVT is built as a sub-project. Whether to use a compiler cache is
+# the top-level project's decision, and CMAKE_CXX_COMPILER_LAUNCHER would apply to all of it.
+function(hvt_setup_compiler_cache)
+    if (NOT ENABLE_COMPILER_CACHE)
```
DDoS: This check doesn't belong here.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+        return()
+    endif()
+
+    if (NOT CMAKE_CXX_COMPILER_LAUNCHER)
+        if (NOT PROJECT_IS_TOP_LEVEL)
+            return()
+        endif()
+
+        find_program(HVT_COMPILER_CACHE NAMES ccache sccache)
+        if (NOT HVT_COMPILER_CACHE)
```
DDoS, line 18: This should output some kind of warning.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+# saves both build time and about a gigabyte of disk in a debug build tree.
+#
+# NOTE: A group must only contain targets that compile with the same preprocessor definitions,
+# because a precompiled header can only be reused by a target whose definitions match the one that
+# produced it. That is why the library and the tests are separate groups: the tests are not built
+# with HVT_BUILD and do add GLEW_STATIC. If a target in a group ever diverges, the compiler
+# rejects the precompiled header and reports which definition differs; give that target its own
+# group, or NO_PRECOMPILED_HEADERS.
+function(hvt_enable_precompiled_header TARGET HEADER GROUP)
+    if (NOT ENABLE_PRECOMPILED_HEADERS)
```
DDoS, line 97: This check doesn't belong here.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+        hvt_enable_precompiled_header(${TARGET} "${_HVT_SOURCE_DIR}/pch.h" sublibrary)
+    endif()
+endfunction()
+
+# Make a target use a precompiled header, if precompiled headers are enabled.
+#
+# Usage:
+#   hvt_enable_precompiled_header(<target> <header> <group>)
+#
+# Targets in the same <group> share one precompiled header: the first one to call this compiles
```
DDoS: > Targets in the same <group> share one precompiled header.

The rest is AI babble.

cmake/HvtHelpers.cmake
```diff
@@ -7,25 +7,135 @@ function(set_if_not_defined VAR_NAME VAR_VALUE VAR_DOC)
...
+# <header>, and every later one reuses that result. Compiling it once rather than once per target
+# saves both build time and about a gigabyte of disk in a debug build tree.
+#
+# NOTE: A group must only contain targets that compile with the same preprocessor definitions,
```
DDoS, line 93: > NOTE: A group must only contain targets that compile with the same preprocessor definitions.

Cut the rest.

cmake/HvtHelpers.cmake
```diff
@@ -58,11 +168,11 @@ function(hvt_embed_sublibraries TARGET)
...
     else()
-        # Objects must be SOURCES, not link items: a static archive has no link line, so
-        # $<TARGET_OBJECTS> given to target_link_libraries() never reaches it. Keep it PRIVATE:
-        # PUBLIC/INTERFACE would leak $<TARGET_OBJECTS> into the installed export.
+        # Bake the private sub-libraries' object files directly into the parent target.
+        # NOTE: The precompiled header has to be filtered out.
         foreach(_lib ${ARGN})
-            target_sources(${TARGET} PRIVATE "$<TARGET_OBJECTS:${_lib}>")
+            target_sources(${TARGET} PRIVATE
+                "$<BUILD_INTERFACE:$<FILTER:$<TARGET_OBJECTS:${_lib}>,EXCLUDE,\\.gch>>")
```
DDoS, line 148: This looks incredibly fragile.

docs/build.md
```diff
@@ -86,6 +86,54 @@ When asked to build and fix until success:
 If **configure** fails, see Build troubleshooting below — most first-time failures are environment
 or dependency issues, not application source bugs.
 
+## Build speed
```
DDoS: This isn't doc, it's AI notes. If you're not going to write proper doc, then just remove it.

include/hvt/engine/viewportEngine.h
```diff
@@ -113,6 +116,11 @@ struct HVT_API FramePassDescriptor
     TaskCreationOptions taskCreationOptions;
 };
 
+/// A scene delegate owning handle.
+/// \note \c UsdImagingDelegate is only forward declared here, because
```
DDoS: Forward declaring is always good practice, so I don't know why the comment needs to even explain yet alone justify it.

source/utils/CMakeLists.txt
```diff
@@ -37,9 +37,13 @@ if(APPLE)
 endif ()
 
 # Define the library.
+# NOTE: The precompiled header is opted out of here because pathUtils.cpp is compiled as
+# Objective-C++ on Apple, and a C++ precompiled header cannot be used by an Objective-C++
+# translation unit. The sub-library is a single, cheap source file, so there is nothing to gain.
 hvt_add_sublibrary(${_TARGET}
     SOURCES ${_SOURCE_FILES}
     HEADERS ${_HEADER_FILES}
+    NO_PRECOMPILED_HEADERS
```
DDoS:
```suggestion
    NO_PRECOMPILED_HEADERS # No PCH support for objc++
```
Remove the other comment.

test/CMakeLists.txt
```diff
@@ -81,6 +84,11 @@ add_executable(${_TARGET}
     tests/testBoundingBoxSceneIndex.cpp
 )
 
+# Use the shared test precompiled header.
+# NOTE: A group of its own, because the test executable and the test framework library do not
```
DDoS, line 88: This note is useless because it just repeats what the group argument does.

test/RenderingFramework/CMakeLists.txt
```diff
@@ -99,6 +99,10 @@ target_compile_definitions(${_TARGET}
         TEST_DATA_OUTPUT_PATH=${test_data_output_normalized}
 )
 
+# Use the shared test precompiled header.
+# NOTE: A group of its own; see the note in test/CMakeLists.txt.
```
DDoS, line 103: useless, remove

CMakeLists.txt
```diff
@@ -51,6 +51,13 @@ string(COMPARE EQUAL "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_SOURCE_DIR}" _HVT_PR
 if (_HVT_PROJECT_IS_TOP_LEVEL AND APPLE)
     set_if_not_defined(CMAKE_OSX_DEPLOYMENT_TARGET "12.0" "Minimum OSX deployment version")
 
+    # Build for the current host architecture.
+    execute_process(COMMAND uname -m
+        OUTPUT_VARIABLE _HOST_ARCH
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    set_if_not_defined(CMAKE_OSX_ARCHITECTURES "${_HOST_ARCH}" "OSX architecture")
```
DDoS: Useless, that's the default behaviour. What you want to do is turn off universal builds, which might be set externally.

CMakeLists.txt
```diff
@@ -119,9 +126,18 @@ endif()
 include(CMakeDependentOption)
 
 # An option to enable the use of precompiled headers.
-# TODO: Implement precompiled headers.
+# NOTE: Turn this off to check that every source file includes what it uses; a precompiled header
+# hides missing includes. See source/pch.h.
 option(ENABLE_PRECOMPILED_HEADERS "Enable precompiled headers" ON)
 
+# An option to enable a compiler cache (ccache or sccache) when one is installed.
+# NOTE: This does not speed up a clean build, but makes rebuilds and branch switches far cheaper.
```
DDoS: Useless: that's the definition of caching...

CMakeLists.txt
```diff
@@ -119,9 +126,18 @@ endif()
 include(CMakeDependentOption)
 
 # An option to enable the use of precompiled headers.
-# TODO: Implement precompiled headers.
+# NOTE: Turn this off to check that every source file includes what it uses; a precompiled header
+# hides missing includes. See source/pch.h.
```
DDoS: > See source/pch.h.

See what? That's just the header.

CMakeLists.txt
```diff
@@ -119,9 +126,18 @@ endif()
...
 option(ENABLE_PRECOMPILED_HEADERS "Enable precompiled headers" ON)
 
+# An option to enable a compiler cache (ccache or sccache) when one is installed.
+# NOTE: This does not speed up a clean build, but makes rebuilds and branch switches far cheaper.
+option(ENABLE_COMPILER_CACHE "Use a compiler cache (ccache/sccache) if one is available" OFF)
+
+# An option to emit limited rather than standalone debug information in the optimized builds.
+# NOTE: This only affects RelWithDebInfo, which is what the asan and ubsan presets build on.
```
DDoS: That just repeats the first line. Might as well rewrite that one as:
> An option to emit limited rather than standalone debug information in the RelWithDebInfo builds. 

This could use an explanation of what "limited rather than standalone" actually means, and why'd you want this.

CMakeLists.txt
```diff
@@ -272,19 +288,19 @@ if (MSVC)
         add_compile_options(/WX)
     endif()
 
-    # Tell compiler to store the debugging information in a PDB for the OBJ file.
-    add_compile_options(/Zi)
+    # Store the debugging information in the object files (/Z7) rather than in a separate compile
+    # PDB (/Zi). The linker still produces a full PDB, so this costs nothing but object file size.
```
DDoS, line 282: Say the why, not the what.

CMakeLists.txt
```diff
@@ -317,6 +333,21 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES "[C|c]lang" OR CMAKE_CXX_COMPILER_ID MATCHE
         endif()
     endif()
 
+    # Apple's Clang defaults to standalone debug info, which emits a full definition for every
```
DDoS: > Full debug information on Clang adds a lot to the link time. Since it's not as necessary for release builds, we gain quite a bit of build speed by enabling Clang's debug information size optimizations (`-fstandalone-debug` turns them off).

That's all you need.

CMakeLists.txt
```diff
@@ -343,6 +374,12 @@ endif()
 
 #---------------------------------------------------------------------------------------------------
 
+# Route compilation through a compiler cache when one is installed. This must happen before any
+# target is defined, because CMAKE_CXX_COMPILER_LAUNCHER is read when a target is created.
+hvt_setup_compiler_cache()
```
DDoS: put the option check here instead

CMakeLists.txt
```diff
@@ -376,6 +413,22 @@ else()
...
 endif()
 message(STATUS "  Warnings as errors       : ${ENABLE_WARNINGS_AS_ERRORS}")
+message(STATUS "  Precompiled headers      : ${ENABLE_PRECOMPILED_HEADERS}")
+if (CMAKE_BUILD_TYPE STREQUAL RelWithDebInfo OR _MULTI_CONFIG)
+    message(STATUS "  Limited debug info       : ${ENABLE_LIMITED_DEBUG_INFO} (RelWithDebInfo only)")
+endif()
+if (NOT ENABLE_COMPILER_CACHE)
+    message(STATUS "  Compiler cache           : OFF")
+elseif (CMAKE_CXX_COMPILER_LAUNCHER)
+    # The launcher may carry environment settings, so report the cache program on its own.
```
DDoS: Those environment settings are still probably good to know.

CMakeLists.txt
```diff
@@ -376,6 +413,22 @@ else()
...
+    if (HVT_COMPILER_CACHE)
+        message(STATUS "  Compiler cache           : ${HVT_COMPILER_CACHE}")
+    else()
+        message(STATUS "  Compiler cache           : ${CMAKE_CXX_COMPILER_LAUNCHER}")
+    endif()
+else()
+    message(STATUS "  Compiler cache           : none (install ccache to speed up rebuilds)")
```
DDoS:
```suggestion
    message(STATUS "  Compiler cache           : <no supported compiler cache found>")
```

CMakePresets.json
```diff
@@ -15,7 +15,8 @@
         "ENABLE_TESTS": "ON",
         "ENABLE_BENCHMARKS": "OFF",
         "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/${presetName}",
-        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON"
+        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
+        "ENABLE_COMPILER_CACHE": "ON"
```
DDoS, line 19:
```suggestion
        "ENABLE_COMPILER_CACHE": "OFF"
```

DDoS (changes requested):

cmake/HvtHelpers.cmake
hodoulp: Cannot test for now so I will let Windows dev to handle that.

cmake/HvtHelpers.cmake
hodoulp: unfortunately, needed arguments with no command line arg equivalent.

cmake/HvtHelpers.cmake
hodoulp: the helper is now self-contain. I prefer that

DDoS: It's not self-contained if it affects all the build commands CMake will issue.

cmake/HvtHelpers.cmake
DDoS: You don't need `env` to add environment variables to a command line.

hodoulp:
ok, that's time to wrap up the work; otherwise, I will never submit anything.

cmake/HvtHelpers.cmake
hodoulp: Finally, using `cmake -E` to make it platform agnostic.

cmake/HvtHelpers.cmake
hodoulp: Anything needed to put in place `ccache` is here and if not possible nothing is done. Devs only have to use `hvt_setup_compiler_cache()` to put in place `ccache`.

cmake/HvtHelpers.cmake
hodoulp: finally using `cmake -E` to make it platform nagnostic.

source/utils/CMakeLists.txt
```diff
@@ -37,9 +37,13 @@ if(APPLE)
 endif ()
 
 # Define the library.
+# NOTE: The precompiled header is opted out of here because pathUtils.cpp is compiled as
+# Objective-C++ on Apple, and a C++ precompiled header cannot be used by an Objective-C++
+# translation unit. The sub-library is a single, cheap source file, so there is nothing to gain.
```
DDoS: Comment is not needed.

DDoS (approved):

PR #199, opened 2026-08-28
AGP-755: upgrade usd to 26.8

wingfiring:
### Changes Made

- 
- 
- 

### Test Configuration
- **Platform(s) tested**: 
- **Build configuration(s)**: 
- **Render delegate(s)**: 
- **OpenUSD version**: 

### Tests Performed

- [ ] Existing unit tests pass

## Documentation

- [ ] Code is self-documenting / well-commented

## Checklist

- [ ] My code follows the project's coding standards
- [ ] My changes generate no new warnings or errors

### DIFF
--- a/source/engine/renderIndexProxy.cpp
+++ b/source/engine/renderIndexProxy.cpp
@@ -25,11 +25,15 @@
 #endif
 // clang-format on
 
+#include <pxr/base/tf/diagnostic.h>
 #include <pxr/imaging/hd/driver.h>
 #include <pxr/imaging/hdSt/renderDelegate.h>
 #include <pxr/imaging/hdSt/resourceRegistry.h>
+
+#if defined(ADSK_OPENUSD_PENDING) && PXR_VERSION < 2608
 #include <pxr/imaging/hgi/hgi.h>
 #include <pxr/imaging/hgi/tokens.h>
+#endif
 
 #if defined(__clang__)
 #pragma clang diagnostic pop
@@ -44,8 +48,12 @@ namespace HVT_NS
 
 RenderIndexProxy::RenderIndexProxy(const std::string& rendererName, HdDriver* hgiDriver)
 {
+    HdRendererPluginRegistry& registry = HdRendererPluginRegistry::GetInstance();
+
+#if defined(ADSK_OPENUSD_PENDING) && PXR_VERSION < 2608
+    // Prior to USD 26.08, the Hgi must be passed to the render delegate through
+    // HdRendererCreateArgs in the render settings map.
     HdRenderSettingsMap settingsMap;
-#ifdef ADSK_OPENUSD_PENDING
     Hgi* hgi = hgiDriver ? hgiDriver->driver.GetWithDefault<Hgi*>() : nullptr;
     if (hgi && hgiDriver->name == HgiTokens->renderDriver)
     {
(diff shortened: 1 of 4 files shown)


hodoulp:
⚠️ You merge two completely different works: SharedContext & 26.08 support in the same PR. But SharedContext is urgent (`HVTEx` master fails), the other is not.

For the SharedContext, why not making it the default behaviour for OpenGL? any concerns? potential limitations?

source/engine/renderIndexProxy.cpp
```diff
@@ -44,23 +43,26 @@ namespace HVT_NS
 
 RenderIndexProxy::RenderIndexProxy(const std::string& rendererName, HdDriver* hgiDriver)
 {
-    HdRenderSettingsMap settingsMap;
-#ifdef ADSK_OPENUSD_PENDING
-    Hgi* hgi = hgiDriver ? hgiDriver->driver.GetWithDefault<Hgi*>() : nullptr;
-    if (hgi && hgiDriver->name == HgiTokens->renderDriver)
+    // Note: the Hgi is handed to the render delegate through the HdDriver passed to
+    // HdRenderIndex::New() below, which forwards it via HdRenderDelegate::SetDrivers().
+    HdRendererPluginRegistry& registry = HdRendererPluginRegistry::GetInstance();
```
hodoulp: The library still needs to support USD 24.11 (i.e., MayaHydra) so I assume a USD version check is more appropriate.

wingfiring:
> ⚠️ You merge two completely different works: SharedContext & 26.08 support in the same PR. But SharedContext is urgent (`HVTEx` master fails), the other is not.
> 
> For the SharedContext, why not making it the default behaviour for OpenGL? any concerns? potential limitations?

You're right — I've split it. The SharedGLContext change is now on its own branch (sunjac/shared-gl-context) off main, separate from the USD 26.08 upgrade. The PR is https://github.com/Autodesk/hydra-viewport-toolbox/pull/200

On making it the default for OpenGL: it's feasible. SharedGLContext removes the infrastructure blocker that the comment in TestHelpers.cpp referred to as "a separate, larger change" — the process-global GL context now exists, sharedHgiCache and the gtest listener are already wired, and OpenGLRendererContext's coupling to its own OpenGLWindow is minimal (the window is never presented to, never resized, and readback goes through Hgi, not the default framebuffer). The code change would be small: add OpenGL to isShareableBackend(), drop the per-test _glWindow usage, add a GL state reset in beginGL().

The concern is test isolation. With a shared context, residual GL state (blend mode, depth mask, bound program, etc.) carries over between tests, and leaked GPU resources accumulate instead of being freed when the per-test context is destroyed. That requires auditing the OpenGL test suite, which is non-trivial but not an architectural blocker. I'd suggest doing it as a follow-up rather than rushing it into this urgent fix.

source/engine/renderIndexProxy.cpp
wingfiring: Fixed.

wingfiring:
the upstream vcpkg PR has been merged.

source/engine/renderIndexProxy.cpp
```diff
@@ -54,13 +62,28 @@ RenderIndexProxy::RenderIndexProxy(const std::string& rendererName, HdDriver* hg
...
     }
+    _renderDelegate = registry.CreateRenderDelegate(TfToken(rendererName), settingsMap);
+#else
+    // Note: the Hgi is handed to the render delegate through the HdDriver passed to
+    // HdRenderIndex::New() below, which forwards it via HdRenderDelegate::SetDrivers().
+    _renderDelegate = registry.CreateRenderDelegate(TfToken(rendererName));
 #endif
+    if (!_renderDelegate)
+    {
+        // A renderer plugin reports itself unsupported by creating the platform default Hgi and
```
DDoS: > reports itself unsupported by creating the platform default Hgi and querying its capabilities

I'm not sure I understand what that means.

source/engine/viewportEngine.cpp
```diff
@@ -237,13 +237,16 @@ void UpdateRendererSettings(RenderIndexProxy* renderIndex)
...
+    if (renderIndex && renderIndex->RenderIndex())
         renderIndex->RenderIndex()->GetRenderDelegate()->Stop();
 
     // Recreate the render index
     renderIndex = RenderIndexProxyPtr(new RenderIndexProxy(desc.rendererName, desc.hgiDriver));
 
-    if (renderIndex)
+    // RenderIndexProxy leaves the render index null when the render delegate could not be created
+    // (it reports the reason); leave the caller with a proxy without a render index rather than
+    // dereferencing it here.
```
DDoS: I don't think we need a three line comment for a null check...

DDoS (approved):

source/engine/renderIndexProxy.cpp
wingfiring: Good point — the original wording was too terse. I reworded it in 9265e80 to spell out the mechanism: `CreateRenderDelegate` returns null when the plugin's `IsSupported()` check fails. For Storm, `IsSupported()` probes the platform default graphics backend (HgiGL on Linux/Windows), whose initialization needs a current OpenGL context — without one the probe fails and the plugin reports itself as unsupported.

DDoS: I still don't see the point here. What's special about HgiGL here? All backends can fail with unsupported to various reasons.
This smells a lot like overly specific AI comments about some issue it encountered. Just remove that comment all together.

wingfiring: You're right — that comment was too specific and didn't belong here. I've removed it.
About the earlier reply: That response was posted by an automated agent without me noticing. Sorry about the confusion.

The remaining change in this PR is just defensive: we log a clear error and return early instead of dereferencing a null pointer. 
The underlying issue we hit in the test suite — Storm's support probe running without a current OpenGL context when Vulkan tests run first — is fixed separately in #200 (SharedGLContext keeps a process-wide GL context current for the probe). That PR is the actual fix; this one only adds the guard.

hodoulp:
Next time, do not forget to update the description of the PR

PR #200, opened 2026-08-28
Keep a shared OpenGL context current for Vulkan tests

wingfiring:
OpenUSD answers "is Storm supported" by building the platform default Hgi (HgiGL on Windows) and reading its capabilities, and it runs that probe without the caller's Hgi. With no OpenGL context current the probe reports an API version of 0, so Storm is declared unsupported even though the test renders with Vulkan, and CreateRenderDelegate() returns null. The first such probe also latches an empty OpenGL entry point table process-wide (GarchGLApiLoad() runs only once), which is why every OpenGL test after the first Vulkan test would fail as well.

Add a SharedGLContext utility to the RenderingFramework that creates a hidden SDL window with an OpenGL context and keeps it current via a gtest event listener that restores it before each test, because the per-test contexts are destroyed when a test ends.

TestHelpers.cpp previously documented this gap as "a separate, larger change" (OpenGL was excluded from the shared-Hgi cache because HgiGL's cached GPU objects live inside the GL context).

### Changes Made

- 
- 
- 

### Test Configuration
- **Platform(s) tested**: 
- **Build configuration(s)**: 
- **Render delegate(s)**: 
- **OpenUSD version**: 

### Tests Performed

- [ ] Existing unit tests pass

## Documentation

- [ ] Code is self-documenting / well-commented

## Checklist

- [ ] My code follows the project's coding standards
- [ ] My changes generate no new warnings or errors

### DIFF
--- a/test/main.cpp
+++ b/test/main.cpp
@@ -14,11 +14,28 @@
 
 #include <gtest/gtest.h>
 
+#include <RenderingFramework/OpenGLTestContext.h>
 #include <RenderingFramework/TestHelpers.h>
 #include <RenderingFramework/UsdHelpers.h>
 #include <RenderingFramework/TestFlags.h>
 
 #include <SDL2/SDL.h>
+
+namespace
+{
+
+/// Restores the shared OpenGL context before each test, because the per-test contexts are
+/// destroyed when a test ends.
+class SharedGLContextListener : public ::testing::EmptyTestEventListener
+{
+public:
+    void OnTestStart(const ::testing::TestInfo&) override
+    {
+        TestHelpers::OpenGLWindow::makeSharedCurrent();
+    }
+};
+
+} // anonymous namespace
 
 int main(int argc, char** argv)
 {
@@ -30,6 +47,18 @@ int main(int argc, char** argv)
     {
         std::cerr << "SDL initialization failed: " << SDL_GetError() << std::endl;
         return EXIT_FAILURE;
+    }
+
+    // Keep a shared OpenGL context current for the whole process so that OpenUSD's Storm
+    // support probe (which reads HgiGL capabilities) succeeds even during Vulkan-only tests.
+    if (TestHelpers::OpenGLWindow::createShared())
+    {
+        ::testing::UnitTest::GetInstance()->listeners().Append(new SharedGLContextListener);
+    }
+    else
+    {
+        std::cerr << "No shared OpenGL context: the Vulkan tests are expected to fail."
+                  << std::endl;
     }
 
     int ret = EXIT_FAILURE;
@@ -50,6 +79,7 @@ int main(int argc, char** argv)
     // and static destructors would only run after main() returns.
     TestHelpers::releaseSharedHgi();
 
+    TestHelpers::OpenGLWindow::destroyShared();
     SDL_Quit();
 
     return ret;
--- a/test/main.cpp
+++ b/test/main.cpp
@@ -14,6 +14,8 @@
 
 #include <gtest/gtest.h>
 
+#include <iostream>
+
 #include <RenderingFramework/OpenGLTestContext.h>
 #include <RenderingFramework/TestHelpers.h>
 #include <RenderingFramework/UsdHelpers.h>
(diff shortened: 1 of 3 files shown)


wingfiring:
On making it the default for OpenGL: it's feasible. SharedGLContext removes the infrastructure blocker that the comment in TestHelpers.cpp referred to as "a separate, larger change" — the process-global GL context now exists, sharedHgiCache and the gtest listener are already wired, and OpenGLRendererContext's coupling to its own OpenGLWindow is minimal (the window is never presented to, never resized, and readback goes through Hgi, not the default framebuffer). The code change would be small: add OpenGL to isShareableBackend(), drop the per-test _glWindow usage, add a GL state reset in beginGL().

The concern is test isolation. With a shared context, residual GL state (blend mode, depth mask, bound program, etc.) carries over between tests, and leaked GPU resources accumulate instead of being freed when the per-test context is destroyed. That requires auditing the OpenGL test suite, which is non-trivial but not an architectural blocker. I'd suggest doing it as a follow-up rather than rushing it into this urgent fix.

test/RenderingFramework/CMakeLists.txt
```diff
@@ -105,6 +107,8 @@ target_link_libraries(${_TARGET}
         hvt
         hd
         $<IF:$<TARGET_EXISTS:GLEW::glew>,GLEW::glew,GLEW::glew_s>
+    PRIVATE
+        GTest::gtest
```
hodoulp: ⚠️ No `gtest` dependency in this library. This library provide helpers to build utest for Metal, OpenGL, etc. backends i.e., independently from the utest machanism.

test/RenderingFramework/SharedGLContext.cpp
```diff
@@ -0,0 +1,128 @@
...
+    if (gSharedWindow && gSharedGLContext && !gListener)
+    {
+        gListener = new SharedGLContextListener();
+        ::testing::UnitTest::GetInstance()->listeners().Append(gListener);
+    }
+}
+
+} // namespace TestHelpers
+
+#endif
```
hodoulp: Still not clearly to understand this class: The OpenGLContext already exists so why not enhancing the existing class?

hodoulp:
> The concern is test isolation. With a shared context, residual GL state (blend mode, depth mask, bound program, etc.) carries over between tests, and leaked GPU resources accumulate instead of being freed when the per-test context is destroyed. That requires auditing the OpenGL test suite, which is non-trivial but not an architectural blocker. I'd suggest doing it as a follow-up rather than rushing it into this urgent fix.

You are right, that's major challenge. @sebastienberube-adsk what was your idea on that subject?

test/RenderingFramework/CMakeLists.txt
wingfiring: Removed. Thanks.

test/RenderingFramework/SharedGLContext.cpp
wingfiring: Simply copied from previous PR... my bad.
Fixed.

test/main.cpp
```diff
@@ -32,6 +49,18 @@ int main(int argc, char** argv)
...
+    if (TestHelpers::OpenGLWindow::createShared())
+    {
+        ::testing::UnitTest::GetInstance()->listeners().Append(new SharedGLContextListener);
+    }
+    else
+    {
+        std::cerr << "No shared OpenGL context: the Vulkan tests are expected to fail."
+                  << std::endl;
+    }
+
```
hodoulp, line 65: ok, I now understand the idea, and that's great. However, the main.cpp is also used for other backends (e.g., Metal). What should be done?

wingfiring: main.cpp is built on macOS too, where OpenGLTestContext is compiled (macOS supports OpenGL). The shared GL context is needed regardless of which backend a test renders with: OpenUSD's Storm support probe reads HgiGL capabilities, which requires a current OpenGL context — even when the test itself uses Metal. So OpenGLWindow::createShared() should be valid on macOS.

hodoulp: That's true.

hodoulp (approved):

PR #194, opened 2026-08-13
Clarify agent docs and public API documentation

hodoulp:
Sync agent-facing documentation and public API header comments with post-merge reality after OutlineManager, dual SI/SD TaskBackend, and related merges. No runtime behavior changes — documentation only.

### Changes Made

- **Agent docs:** Update `AGENTS.md`, `README.md`, `test/README.md`, `docs/ai-agents.md`, and `.github/copilot-instructions.md` — howTo21 numbering, SI/SD architecture, `HVT_API` on public classes, CI pointer fixes, Cursor `.cursorignore` / `.cursorindexingignore` split
- **`test/README.md`:** Fix scene-index API examples (`BoundingBoxSceneIndex`, etc.), add HowTo21 / `OutlineManager` section
- **`UpdatePrim`:** Correct and expand Doxygen in `viewportEngine.h` (param order, xform-op requirements, scale semantics, child visibility)
- **`enableOutline`:** Disambiguate legacy Hdx tasks vs GPU `OutlineManager` in `basicLayerParams.h`, `selectionSettingsProvider.h`, `outlineManager.h`
- **`FramePass`:** Document SI/SD `TaskBackend` selection; link to `taskBackend.h` and `docs/framepass.md`; document `HVT_USE_LEGACY_SCENE_DELEGATE` in `docs/framepass.md`
- **Flash picking:** Document Flash vs Storm pick wiring in `flashPickTask.h` and `CreateFlashPickTask`
- **`CreateUSDSceneIndex`:** Clarify overload purposes in `viewportEngine.h`
- **`include/hvt/api.h`:** Align export macro comment with public-class `HVT_API` requirement and scene-index exception
- **`docs/README.md`:** Add flash-picking row; note `UpdatePrim` on viewport engine entry

## Testing

Documentation-only change — no code behavior modified. Verified paths and API names referenced in updated docs against current headers and how-tos.

### Test Configuration
- **Platform(s) tested**: N/A (docs only)
- **Build configuration(s)**: N/A
- **Render delegate(s)**: N/A
- **OpenUSD version**: N/A

### Tests Performed
- [x] Existing unit tests pass (not re-run — no code changes)
- [ ] Added/Updated unit test(s) for the changes
- [ ] Tested on multiple platforms
- [ ] Tested with different render delegates
- [ ] Performance testing (if applicable)

## Documentation

- [x] Code is self-documenting / well-commented
- [x] Public API changes are documented with Doxygen comments

## Checklist

[...]

PR #204, opened 2026-09-09
Fix vcpkg NuGet binary cache auth on self-hosted runner

laforel:
### Timeline
- **May-Jun 2026**: CI builds took ~10 min (vcpkg NuGet cache working: `Restored 45 package(s) from NuGet in 2.4 min`)
- **Aug 12, 2026** (PR #186, USD 26.5): Last fast build. NuGet cache working.
- **Aug 31, 2026** (PR #199, USD 26.8 upgrade): NuGet cache broke. `Restored 0 package(s)` + `credential providers failed to authenticate`. Build took 1107 min.
- **Sep 2-8, 2026**: All subsequent builds take 80-90+ min (rebuilding USD from source every time).

### What broke
PR #199 (USD 26.8 upgrade) also bumped the vcpkg submodule from `d592849` to `6c409c8` (842 commits). The newer vcpkg version downloads nuget-7.9.0 (was 7.8.x). This newer nuget.exe on Linux/mono fails to authenticate with GitHub Packages:
- **Restore**: `Restored 0 package(s) from NuGet` — zero packages downloaded, full rebuild every time
- **Upload**: `One or more NuGet credential providers failed to authenticate` — cache never populated

### Impact
Every CI run rebuilds USD from source (~60-90 min) instead of downloading from cache (~2 min). A docs-only PR (#194) took 24 hours because of this.

### Fix: Local file cache for self-hosted GPU runner

Since the NuGet credential provider is broken on Linux/mono with the newer vcpkg, this PR adds a **local file cache** for the self-hosted GPU runner (inspired by sunjac's `persist-vcpkg-cache` branch):

- The Configure step now sets up `~/.hvt-vcpkg-installed` outside the checkout (survives `clean: true`)
- Hashes `vcpkg.json` to detect version changes — clears cache only when dependencies change
- Passes `-DVCPKG_INSTALLED_DIR` to CMake so vcpkg uses the persisted directory
- **First build after a USD version change will be slow (~80 min), but subsequent builds will be fast (~10 min)**
- Also adds `packages: write` permission to the workflow for when the NuGet auth is eventually fixed

### Status
- The local cache is working (verified: cmake command has `-DVCPKG_INSTALLED_DIR=/home/gharunner/.hvt-vcpkg-installed`)
- The first build with this PR will be slow (building USD from source, cache was cleared)
- **The next build after this one merges will be fast** — the installed dir will persist on the VM

### Patrick's PRs (internal repo)
- **PR #331** (AGP-832): Fix Linux UT not starting — `docker run -u agpdevops` fails because user doesn't exist in the builder image. Replaces with `--user $(id -u):$(id -g)` and `HOME=/tmp`. Also replaces `|| true` with exit-code accumulator. **Looks correct.**
- **PR #332** (AGP-833): Same `|| true` masking fix for macOS scripts. Uses `pipestatus` for iOS. **Looks correct.**

Both PRs are small, focused CI script fixes. The test masking bug (Linux tests reporting SUCCESS while running zero tests) is real and well-diagnosed.

### DIFF
--- a/.github/actions/setup-vcpkg-cache/action.yaml
+++ b/.github/actions/setup-vcpkg-cache/action.yaml
@@ -63,4 +63,24 @@ runs:
     - name: Set binary caching env var
       shell: bash
       run: |
-        echo "VCPKG_BINARY_SOURCES=clear;nuget,https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json,readwrite" >> $GITHUB_ENV
+        # Write a nuget.config with credentials that vcpkg's internal
+        # nuget.exe invocation will pick up. Using a config file path
+        # (instead of a URL) in VCPKG_BINARY_SOURCES makes nuget.exe
+        # use the <packageSourceCredentials> section for auth.
+        NUGET_CONFIG="$HOME/.nuget/NuGet/NuGet.Config"
+        mkdir -p "$(dirname "$NUGET_CONFIG")"
+        cat > "$NUGET_CONFIG" <<EOF
+        <?xml version="1.0" encoding="utf-8"?>
+        <configuration>
+          <packageSources>
+            <add key="github" value="https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json" />
+          </packageSources>
+          <packageSourceCredentials>
+            <github>
+              <add key="Username" value="${{ github.repository_owner }}" />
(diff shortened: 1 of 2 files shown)


hodoulp (approved):

laforel:
Closing - root cause was an expired PAT (GH_PACKAGES_TOKEN), not the workflow. Token has been regenerated and SSO re-authorized; the existing NuGet cache setup works as-is.

PR #209, opened 2026-09-17
AGP-869 - Reuse the shared GL context for test contexts

hodoulp:
## Description

The pull request changes the code to share the OpenGL context when running unit tests. The most performance benefit is expected to come from the Windows platform based on previous experimentations.

## Explanations

Every test context created an SDL window and a GL context of its own, then destroyed both when the test ended. Measured on macOS that pair costs about **12 ms per test** (~8 ms for `SDL_CreateWindow`, ~4 ms for `SDL_GL_CreateContext`), which dominates per-test setup — creating the `Hgi`, by comparison, takes about **0.1 ms**.

Note: The per-test window is only ever a carrier for a current GL context: a test renders into the frame pass AOVs, which `captureColorTexture()` reads back through `Hgi`, and never draws to the default framebuffer or calls `swapBuffers()`. So the window's size and pixel format cannot affect a result, which means the process-wide window and context that `main()` already creates for OpenUSD's Storm capability probe can back every test context.

### Changes Made

- `OpenGLWindow` borrows the process-wide window/context created by `createShared()` instead of creating its own pair, leaving ownership with `createShared()`/`destroyShared()`.
- Falls back to a dedicated window and context when no shared one exists (e.g. `createShared()` failed on a host without a display).
- Opt out with `HVT_TEST_SHARE_GL_CONTEXT=0` when debugging a test sensitive to residual GL state, mirroring the existing `HVT_TEST_SHARE_HGI` switch.
- Corrects the `sharedHgiCache` comment in `TestHelpers.cpp`, which claimed OpenGL was excluded because "OpenGL context/shader setup is already cheap" (it is not), and implied the cache applies on desktop — `OpenGLRendererContext` passes no backend token, so it never consults the cache.

### Testing

Use `start=$SECONDS; ./build/release/bin/hvt_test; end=$SECONDS; echo "time => $(($end-$start))"` to test.

On macOS, the speed up is not expected to be significant.

## Tests Performed

## Documentation

## Checklist

- [x] **I have signed the Contributor License Agreement (CLA)**

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wingfiring (approved):
LGTM

